PR #30, opened 2016-07-01
Very minimal main function

Luthaf:
This add a very minimal binary to cymbalum, using the recent input module to setup the run. 

The main function should be expanded with a real interface and better error messages, but this is a starting point, allowing people to run simulations without writing code!


Luthaf:
@g-bauer, any comment/suggestion on that?

g-bauer:
I am not familiar with the input module yet, but as you mention, this seems to be a good starting point.

Issue #28, opened 2016-06-27
Add computation for truncated potentials with tail corrections  


g-bauer:
We should add a struct and computation for truncated intermolecular pair potentials with long range corrections.

In Monte-Carlo (MC) simulations we often use potential functions that are truncated at a cutoff distance `rc`. Since we don't have to evaluate forces in MC, we don't have to shift the potential at the cutoff distance. 

Under the **assumption** that the **pair correlation function is unity beyond `rc`**, we can compute a **tail correction** (for the energy and the pressure) by integration of the pair-potential from `rc` to infinity. For some potentials this can be solved analytically. The tail correction is a function of the cutoff distance `rc` and the ~~systems' density~~ densities of all components (`ParticleKind`).

Luthaf:
I believe this would be better as two additional functions for `PairPotential` trait: 

```rust
trait PairPotential {
   // Default return values is equivalent to today's code
   fn tail_energy(&self, cutoff: f64, density: f64) -> f64 {0.0}
   fn tail_virial(&self, cutoff: f64, density: f64) -> Matrix3 {Matrix3::zero()}
}
```

These function would be called from `System::{energy,virial}` with the actual density of the system, and the half of the lower cell parameter as default cutoff. (Is this correct for non-cubic boxes?)

Then, the `CutoffComputation` can implement these by overwriting the cutoff value.
```rust
impl<T: PairPotential> PairPotential for CutoffComputation<T> {
   fn tail_energy(&self, cutoff: f64, density: f64) -> f64 {
        assert!(cutoff >= self.cutoff, "Cutoff is too big in CutoffComputation");
        self.potential.tail_energy(self.cutoff, density)
   }

   fn tail_virial(&self, cutoff: f64, density: f64) -> Matrix3 {
        assert!(cutoff >= self.cutoff, "Cutoff is too big in CutoffComputation");
        self.potential.tail_virial(self.cutoff, density)
   }
}
```

g-bauer:
That's also possible. Note that potentials that are computed via `CutoffComputation` must not be shifted when cutoff corrections are applied.

Luthaf:
Forgot about this one. Maybe there should be another `Computation`? Like having `CutoffComputation` which only uses a cutoff for MC (but is not shifted), and `ShiftedComputation` which uses a cutoff and shift the potential for MD.

g-bauer:
[...]

PR #34, opened 2016-07-13
Refactor pairs interactions

Luthaf:
This is preliminary work for #28, and not finished yet. Are missing:

- [x] A way to provide the cutoff value and the shifted/not-shifted behaviour in the input;
- [x] Documentation;
- [x] More integration tests;

I am opening this to get a feedback on the API. @g-bauer, this will allow to add tail corrections by:

1. adding a `Tail` variant to [PairComputation](https://github.com/Luthaf/cymbalum/blob/e5fd14ab1e7d6795e3c9251072bd9121046abb6b/src/potentials/pairs.rs#L8); 
2. adding the `tail_energy` and `tail_virial` methods to `PairPotential`

An unresolved question is what to do with the `TableComputation`. This make `CutoffComputation` useless, and we may want to unify the TableComputation in the `PairComputation` enum. But we may also want to use table interpolations for other bonded interactions (bond, angles, ...)

### DIFF
--- a/doc/src/SUMMARY.md
+++ b/doc/src/SUMMARY.md
@@ -12,7 +12,7 @@
     - [Adsorption in zeolites]()
 
 
-- [Input files](input/intro.md)
+- [Input files reference](input/intro.md)
     - [Systems](input/systems.md)
     - [Interactions](input/interactions.md)
     - [Potentials](input/potentials.md)
--- a/doc/src/input/interactions.md
+++ b/doc/src/input/interactions.md
@@ -15,6 +15,9 @@ An example of an input file for the f-SPC model of water is given bellow:
 [input]
 version = 1
 
+[global]
+cutoff = "12.5 A"
+
 [[pairs]]
 atoms = ["O", "O"]
 lj = {sigma = "3.16 A", epsilon = "0.155 kcal/mol"}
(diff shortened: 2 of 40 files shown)


Luthaf:
> You can shift the potential to zero and add another contribution so that the forces are smooth and continuous. (from #28)

Also, this kind of smoothing function can be added as another variant for the `PairComputation`

g-bauer:
> An unresolved question is what to do with the `TableComputation`. This make `CutoffComputation` useless, and we may want to unify the `TableComputation` in the `PairComputation` enum.

If I am understanding the concept of `Computation` right, it should encapsulate different ways of evaluating the `Potential`s `energy` and `force`. For example, a `TableComputation` yields values for energies and forces via interpolation from tables.

The `PairComputation` like you implemented it is more of a `CutoffTreatment` than a `Computation`, right? (I hope you get what I am trying to say with this) A tabulated potential could be treated with the same `PairComputation`, i.e. one could apply `cutoff`, `shift`, `tail_correction` and therefore I am not sure if 

>  we may want to unify the TableComputation in the PairComputation enum

would make sense. I am not sure, where to add the `Computation`, maybe inside `PairInteraction`?

```rust
/// Different ways to treat potentials that are truncated at a cutoff distance
pub struct CutoffTreatment {
    /// Truncated potential without corrections or shifts
    None, // or Cutoff,
    /// Truncated and shifted potential
    /// such that the potential at cutoff is zero
    Shift(f64),
    /// Truncated and shifted potential using a switch function
    /// to have smooth and continuous forces
    SmoothShift(f64),
    /// Truncated potential with tail correction,
    /// assuming the pair distribution function to be unity for r > cutoff
    TailCorrection, 
}

pub struct PairInteraction {
    /// The potential of this interaction
    potential: Box<PairPotential>,
    /// The associated pair restrictions
    restriction: PairRestriction,
    /// The cutoff distance
    cutoff: f64,
    /// The way to treat potential truncation
    /// before: computation: PairComputation
    cutoff_treatment: CutoffTreatment,
    /// The way to compute the energies and forces
    computation: PairComputation
}
```

Just an idea, I am not sure if this is feasible.

g-bauer:
> A way to provide the cutoff value and the shifted/not-shifted behaviour in the input;

I think it makes sense to treat this separately from the pair potentials, i.e. to set the behavior once for all pair potentials. 

Maybe it is even a good idea to set `cutoff` once and not for every `PairInteraction`?

Luthaf:
> I think it makes sense to treat this separately from the pair potentials, i.e. to set the behavior once for all pair potentials. 

I like the ability to use different cutoff for different interactions, it can be useful sometimes (using both harmonic and Lennard-Jones interactions is the first example that come in my mind). But it could be nice to have a way to specify a default cutoff, which would be used if no other is given.

Luthaf:
This is good to merge for me, follow up work can come to add tail correction and smoothed potential. @g-bauer, do you have comments on this before I merge?

PR #44, opened 2016-10-11
Pressure computation and Monte-Carlo

Luthaf:
I went for an easy solution, asking the propagator for an external temperature strategy and using that for pressure/stress computations inside the system.


g-bauer:
I like it. 👍

PR #46, opened 2016-10-22
Less public names, disable glob imports

Luthaf:
The main change of this PR is the removal of `pub use <module>::*;` from lumol_core, preventing usage of lumol as `use lumol::*;`. Because all the names must now be fully qualified, it also rename the `simulation` and `system` modules to the shorter `sim` and `sys` respectively.

This is again a massive breaking change; I try to do most of the one I have been thinking about before hitting 0.1

Break all the things!


Luthaf:
@g-bauer are you comfortable with me merging that, or do you prefer to wait for your changes to land in master?

g-bauer:
I'm fine with it. Go ahead and merge. I will modify my changes accordingly.

PR #48, opened 2016-10-26
fixed bug with inverse temperature in cost function...

g-bauer:
... for rotation and translation moves.


Luthaf:
\o/ Thank you for the PR!

When Travis is OK, I'll merge this.

Issue #23, opened 2016-03-26
Monte Carlo moves: adjustable maximum displacement


g-bauer:
It would be a good idea to implement a `demanded acceptance` for MC moves that have a displacement (translation, volume or box change). That way, the maximum displacement for the move can be adjusted during the simulation to achieve that acceptance. With a fixed displacement as is, simulations can be very inefficient.

Luthaf:
Yes, I agree. This can be the default, but should be easy to deactivate, as modifying the MC parameters will change the thermodynamic ensemble.

I have seen this implemented using something like an `adjust(&mut self, direction: Direction)` method, which could be added to the `Move` trait. The Monte-Carlo propagator then takes care of counting the current acceptance rate, and calling this method with either `More` or `Less` for the wanted adjustment in acceptance rate.

g-bauer:
I am currently working on this. Having displacement adjustments accelerates equilibration by a lot so in order to verify other implementations (like NPT, tail corrections, comparison to NIST data) I think it is nice to have this first.

I will create a PR soon. For the moment, I added a new struct `Displacement` which will store:

- number of times a move was called
- number of times a move was accepted
- target acceptance of a move
- update frequency (0 = no updates = fixed displacement, 100 = after 100 calls to move, update delta)
- the current delta (maximum displacement)
- the current range from which to sample
- a method to update the current delta + range according to the current acceptance

The structure replaces `delta` and `delta_range` inside a move. Also the `McMove` will have a new function `update_displacement` (or smth. like that). Moves that do not need a displacement (insertion, deletion) just do nothing when calling the update function. 

I just wondering how much should be done via functions or by direct manipulation of the struct's components (e.g. how to increase the counter for a move, accepted move etc.).

Does that sound reasonable?

Luthaf:
This seems to be a nice implementation. Another way to do it would be to store an `Update` (this is a very bad name) struct alongside with the Move:

```rust
struct Update {
    called: usize,
    accepted: usize,
    frequency: usize
}

struct MonteCarlo {
    moves: Vec<(Box<Move>, Update)>
}
```

So that it is the responsibility of `MonteCarlo` to keep track of the number of time a move was called. This make  implementation of new moves easier (we do not need to remember to update the move counts), but leave the `delta`/`delta_range` inside the moves.

> I just wondering how much should be done via functions or by direct manipulation of the struct's components (e.g. how to increase the counter for a move, accepted move etc.).

[...]

Issue #16, opened 2016-02-13
Implement Hybrid Monte-Carlo


Luthaf:
This is easy: just use the functionalities around to build the hybrid move

Luthaf:
See http://www.sciencedirect.com/science/article/pii/037026938791197X for the paper describing this method.

Luthaf:
> A question popping into my mind would be: what happens when NVE is performed and temperature drops considerably?
>
> @g-bauer in https://github.com/lumol-org/lumol/issues/51#issuecomment-257723795

I do not remember, but I have seen this. I'll check my notes tomorrow. I think that this is why you will want to use NVT simulation instead of NVE with the same temperature as the MC simulation. If/when the MD simulation is at equilibrium, the actual potential energy should only fluctuate around the MC energy.

PR #52, opened 2016-11-01
Implement feature: adjustable maximum displacement in Monte-Carlo

g-bauer:
This implementation addresses issue #23 .

There is still work to do to make the changes work with the `lumol input` crate.

I'd appreciate feedback/critique/suggestions and questions.

### DIFF
--- a/src/core/src/sim/mc/monte_carlo.rs
+++ b/src/core/src/sim/mc/monte_carlo.rs
@@ -87,12 +87,6 @@ impl MonteCarlo {
                 we can not add new moves."
             );
         }
-        if target_acceptance >= 1.0 || target_acceptance < 0.0 {
-            fatal_error!(
-                "The target acceptance ratio of the move has to be a positive \
-                value smaller equal than 1."
-            );
-        }
         self.moves.push((mcmove, MoveCounter::new(Some(target_acceptance))));
         self.frequencies.push(frequency);
     }
@@ -221,25 +215,36 @@ impl MoveCounter {
     /// Create a new counter for the move, initializing all counts to zero and
     /// setting the `target_acceptance`.
     pub fn new(target_acceptance: Option<f64>) -> MoveCounter {
-        MoveCounter{
+        let mut counter = MoveCounter{
             ncalled: 0,
             naccepted: 0,
             nattempted: 0,
-            target_acceptance: target_acceptance,
-        }
+            target_acceptance: None,
+        };
+        counter.set_acceptance(target_acceptance);
+        counter
     }
 
     /// Set the target acceptance for the move counter.
-    pub fn set_acceptance(&mut self, target_acceptance: f64) {
-        self.target_acceptance = Some(target_acceptance);
+    pub fn set_acceptance(&mut self, target_acceptance: Option<f64>) {
+        // Check if `target_acceptance` has a valid value.
+        if target_acceptance.is_some() {
+            let ta = target_acceptance.unwrap();
+            if ta >= 1.0 || ta < 0.0 {
+                fatal_error!(
+                    "The target acceptance ratio of the move has to be a positive value smaller equal than 1."
+                )
+            }
+        }
+        self.target_acceptance = target_acceptance;
     }
-
+    
     /// Compute a scaling factor according to the desired acceptance.
     pub fn compute_scaling_factor(&self) -> Option<f64> {
         // Check if there exists an target_acceptance
         if let Some(ta) = self.target_acceptance {
             // Capture division by zero
-            if self.nattempted == 0 { return Some(1.0) };
+            if self.nattempted == 0 { return None };
             let quotient = self.naccepted as f64 / self.nattempted as f64 / ta;
             // Limit the change
             match quotient {
@@ -259,7 +264,7 @@ impl Default for MoveCounter {
 
 #[cfg(test)]
 mod tests {
-    use sim::mc::{MonteCarlo, MCMove};
+    use sim::mc::{MonteCarlo, MCMove, MoveCounter};
     use sim::Propagator;
     use sys::{System, EnergyCache};
     use rand::Rng;
@@ -302,5 +307,33 @@ mod tests {
         mc.add(Box::new(DummyMove), 1.0);
         mc.setup(&System::new());
         mc.add(Box::new(DummyMove), 1.0);
+    }
+
+    #[test]
+    #[should_panic]
+    fn invalid_acceptance() {
+        let mut mc = MonteCarlo::new(100.0);
+        mc.add_move_with_acceptance(Box::new(DummyMove), 1.0, 0.5);
+        mc.moves[0].1.set_acceptance(Some(1.1));
+    }
+
+    #[test]
+    fn valid_acceptance() {
+        let mut mc = MonteCarlo::new(100.0);
+        mc.add_move_with_acceptance(Box::new(DummyMove), 1.0, 0.5);
+        assert_eq!(mc.moves[0].1.target_acceptance, Some(0.5));
+        mc.moves[0].1.set_acceptance(None);
+        assert_eq!(mc.moves[0].1.target_acceptance, None);     
+    }
+
+    #[test]
+    fn scaling_factor() {
+        let mut counter = MoveCounter::new(Some(0.5));
+        assert_eq!(counter.compute_scaling_factor(), None);
+        counter.nattempted = 10;
+        counter.naccepted = 10;
+        assert_eq!(counter.compute_scaling_factor(), Some(1.2));
+        counter.naccepted = 0;
+        assert_eq!(counter.compute_scaling_factor(), Some(0.8));
     }
 }
--- a/src/core/src/sim/mc/monte_carlo.rs
+++ b/src/core/src/sim/mc/monte_carlo.rs
@@ -15,9 +15,12 @@ pub struct MonteCarlo {
     /// Boltzmann factor: beta = 1/(kB * T)
     beta: f64,
     /// List of possible Monte-Carlo moves
-    moves: Vec<Box<MCMove>>,
-    /// Cumulative frequencies of the Monte-Carlo moves
+    moves: Vec<(Box<MCMove>, MoveCounter)>,
+    /// Cummulative frequencies of the Monte-Carlo moves
     frequencies: Vec<f64>,
+    /// Specifies the number of moves after which an update of a move's
+    /// amplitude is performed.
+    update_frequency: u64,
     /// Random number generator for the simulation. All random state will be
     /// taken from this.
     rng: Box<rand::Rng>,
@@ -44,13 +47,14 @@ impl MonteCarlo {
             beta: 1.0 / (K_BOLTZMANN * temperature),
             moves: Vec::new(),
             frequencies: Vec::new(),
+            update_frequency: 0,
             rng: rng,
             cache: EnergyCache::new(),
             initialized: false,
         }
     }
 
-    /// Add a the `mcmove` Monte-Carlo move to this propagator, with frequency
+    /// Add the `mcmove` Monte-Carlo move to this propagator, with frequency
     /// `frequency`. All calls to this function should happen before any
     /// simulation run.
     ///
@@ -64,8 +68,39 @@ impl MonteCarlo {
                 we can not add new moves."
             );
         }
-        self.moves.push(mcmove);
+        self.moves.push((mcmove, MoveCounter::new(None)));
         self.frequencies.push(frequency);
+    }
+
+    /// Add the `mcmove` Monte-Carlo move to the propagator.
+    /// `frequency` describes how frequent a move is called, `target_acceptance`
+    /// is the desired acceptance ratio of the move.
+    ///
+    /// # Panics
+    ///
+    /// If called after a simulation run.
+    /// If `target_acceptance` is either negative or larger than one.
+    pub fn add_move_with_acceptance(&mut self, mcmove: Box<MCMove>, frequency: f64, target_acceptance: f64) {
+        if self.initialized {
+            fatal_error!(
+                "Monte-Carlo simulation has already been initialized, \
+                we can not add new moves."
+            );
+        }
+        if target_acceptance >= 1.0 || target_acceptance < 0.0 {
+            fatal_error!(
+                "The target acceptance ratio of the move has to be a positive \
+                value smaller equal than 1."
+            );
+        }
+        self.moves.push((mcmove, MoveCounter::new(Some(target_acceptance))));
+        self.frequencies.push(frequency);
+    }
+
+    /// Set the number of times a move has to be called before its amplitude
+    /// is updated. This value is applied to all moves.
+    pub fn set_amplitude_update_frequency(&mut self, frequency: u64) {
+        self.update_frequency = frequency;
     }
 
     /// Get the temperature of the simulation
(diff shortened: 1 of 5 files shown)


src/core/src/sim/mc/moves/mod.rs
```diff
@@ -65,6 +70,76 @@ fn select_molecule(system: &System, moltype: Option<u64>, rng: &mut Box<Rng>) ->
...
+/// 
+/// We can modify (`update`) the maximum displacement to achieve a target acceptance rate 
+/// of a move based on the running acceptance.
+pub struct Displacement {
+    /// Current maximum displacement
+    delta: f64,
+    /// Current sample range
+    pub delta_range: Range<f64>,
+    /// Number of move calls
+    pub ncalls: u64,
```
Luthaf: What is the difference between `ncall` and `nattempted`?

src/core/src/sim/mc/moves/mod.rs
```diff
@@ -65,6 +70,76 @@ fn select_molecule(system: &System, moltype: Option<u64>, rng: &mut Box<Rng>) ->
...
+/// and how often it was accepted. It also stores the current maximum displacement,
+/// i.e. the magnitude of a particle's translation/rotation or a volume's change in size.
+/// 
+/// We can modify (`update`) the maximum displacement to achieve a target acceptance rate 
+/// of a move based on the running acceptance.
+pub struct Displacement {
+    /// Current maximum displacement
+    delta: f64,
+    /// Current sample range
+    pub delta_range: Range<f64>,
```
Luthaf: this could just be called `range`

src/core/src/sim/mc/moves/mod.rs
```diff
@@ -65,6 +70,76 @@ fn select_molecule(system: &System, moltype: Option<u64>, rng: &mut Box<Rng>) ->
...
+            delta_range: Range::new(-delta, delta),
+            ncalls: 0,
+            naccepted: 0,
+            nattempted: 0,
+            target_acceptance: target_acceptance,
+            update_frequency: update_frequency,
+        }
+    }
+    
+    /// Change `update_frequency` to zero, which will prevent further updates.
```
Luthaf: We should also have a way to change the update frequency (`pub fn update_frequency(&mut self, frequency: usize)`) here.

I would also prefer the documentation not to refer internal implementation details. How about:

> Disable future updates of the move amplitude, until a call to `update_frequency` is done with a non zero value

src/core/src/sim/mc/moves/mod.rs
```diff
@@ -65,6 +70,76 @@ fn select_molecule(system: &System, moltype: Option<u64>, rng: &mut Box<Rng>) ->
...
+            return
+        } 
+        // get current acceptance
+        let current_acceptance = self.naccepted as f64 / self.nattempted as f64;
+        // limit changes:
+        // if quotient exceeds a min/max value of 50%, limit changes to 50%.
+        let quotient = current_acceptance / self.target_acceptance;
+        match quotient {
+            _ if quotient > 1.5 => self.delta *= 1.5,
+            _ if quotient < 0.5 => self.delta *= 0.5,
```
Luthaf: 50% seems to be a very big change. I have seen 10% in other software, do you have a specific reason to pick 50% ?

src/core/src/sim/mc/moves/rotate.rs
```diff
@@ -5,9 +5,11 @@ use rand::Rng;
 
 use std::usize;
 use std::f64;
+use std::f64::consts::{PI};
```
Luthaf: style: no need for brackets here

src/core/src/sim/mc/moves/rotate.rs
```diff
@@ -20,41 +22,38 @@ pub struct Rotate {
...
+    /// Displacement
+    displacement: Displacement,
 }
 
 impl Rotate {
     /// Create a new `Rotate` move, with maximum angular displacement of `theta`,
     /// rotating all the molecules in the system.
-    pub fn new(theta: f64) -> Rotate {
-        Rotate::create(theta, None)
+    pub fn new(theta: f64, update_freq: u64, target_acceptance: f64) -> Rotate {
```
Luthaf: I would not add all the parameters to the constructor, but rather use the default and have setters when deviating from the default.

Something like this:

```rust
pub fn new(theta: f64) -> Rotate {
      // same as before, using Displacement::default for the initial value of displacement
}

pub fn set_target_acceptance(&mut self, target: f64) { /* ... */ }

pub fn set_update_frequency(&mut self, frequency: f64) { /* ... */ }
```

src/core/src/sim/mc/moves/rotate.rs
```diff
@@ -154,3 +151,15 @@ fn rotation_matrix(axis: &Vector3D, angle: f64) -> Matrix3 {
 
     return rotation;
 }
+
+/// Returns a random unit vector,
+/// i.e. a random point on an unit sphere.
+/// @TODO: Maybe move this to types::Vector3D?
+fn random_unit_vector3d (rng: &mut Box<Rng>) -> Vector3D {
```
Luthaf: The algorithm changed here. Why not use the nice 3D normal distribution => uniform distribution on the R3 sphere feature?

src/core/src/sim/mc/moves/mod.rs
g-bauer: `nattempted` (as well as `naccepted`) will be set to zero after an "successful" update. `ncalls` continues to count. Currently, `ncalls` is only used to check if the displacement should be updated (and the information about the total numbers of calls to the move). We could also use it in conjunction with a `nequilibration` variable which automatically switches off updates after a certain number of trials.

We could eliminate `ncalls` and compare the frequency directly to `nattempts` in `update`.

src/core/src/sim/mc/moves/mod.rs
g-bauer: I thought about that as well. The question is, where should the user call the function? After adding the move to `MonteCarlo` and while running the simulation, the user currently has no control over the states of the moves.

I will change the documentation as you requested.

Luthaf:
From the design point of view, I still prefer to have the `MonteCarlo` struct
manage the moves instead of all the moves having a `Displacement` field. Three
reason for that:

- `update` is just a delegation to `displacement.update()` in the current code;
- you forgot to update the displacement counter in `Rotate`, so other will make the same mistake and forget this too;
- for Hybrid MC (#16), the adjustable parameter is not the amplitude of the move, but the number of steps of the MD run (an integer and not a float).

---

I would go for something like the following.

PS: This was going to be small, but turned into a huge pile of code. Ask me if you
do not get what I wrote, and tell me if you disagree!

So we start with a counter for moves frequency:

```rust
struct MoveCounter {
    pub ncalls: u64,
    pub naccepted: u64,
    pub nattempted: u64,
    target_acceptance: f64,
}
```

Then the MCMove win a new method to update itself in a given direction

```rust
pub enum UpdateDirection {
    /// We need the acceptance to raise by the given ratio
    More(f64),
    /// We need the acceptance to lower by the given ratio
    Less(f64),
    /// The acceptance does not need to change
    Same,
}

pub trait MCMove {
    // ...

    fn update_acceptance(&mut self, direction: UpdateDirection);
}

```

Method that can be implemented like this

```rust
impl MCMove for Translate {
    // ...
    fn update_acceptance(&mut self, direction: UpdateDirection) {
        match direction {
            UpdateDirection::Same => {return}
            UpdateDirection::More(mut ratio) => {
                debug_assert!(ratio > 1.0);
                if ratio > 1.5 {ratio = 1.5}
                self.delta *= ratio;
            }
            UpdateDirection::Less(mut ratio) => {
                debug_assert!(ratio < 1.0);
                assert!(ratio > 0.0);
                if ratio < 0.5 {ratio = 0.5}
                self.delta *= ratio;
            }
        }
        self.range = Range(-self.delta, self.delta);
    }
}
```

And finally we can use this in the MonteCarlo struct

```rust
pub struct MonteCarlo {
    // ...
    /// List of possible Monte-Carlo moves and associated counters
    moves: Vec<(Box<MCMove>, MoveCounter)>,
    /// Use the same update frequency for all the moves
    update_frequency: u64,
    /// Number of moves counter
    nmoves: u64,
}

impl Propagator for MonteCarlo {
    fn propagate(&mut self, system: &mut System) {
        let mcmove = {
            let probability = self.rng.next_f64();
            // Get the index of the first move with frequency >= probability.
            let (i, _) = self.frequencies.iter()
                                         .enumerate()
                                         .find(|&(_, f)| probability <= *f)
                                         .expect("Could not find a move in MonteCarlo moves list");
            &mut self.moves[i]
        };

        // Update the counters
        mcmove.1.nattempted += 1;
        self.nmoves += 1;

        trace!("Selected move is '{}'", mcmove.0.describe());

        if !mcmove.0.prepare(system, &mut self.rng) {
            trace!("    --> Can not perform the move");
            return;
        }

        let cost = mcmove.0.cost(system, self.beta, &mut self.cache);
        trace!("    --> Move cost is {}", cost);

        let accepted = cost <= 0.0 || self.rng.next_f64() < f64::exp(-cost);

        if accepted {
            // Update the counters
            mcmove.1.naccepted += 1;
            trace!("    --> Move was accepted");
            mcmove.0.apply(system);
            self.cache.update(system);
        } else {
            trace!("    --> Move was rejected");
            mcmove.0.restore(system);
        }

        // Do the adjustments in moves amplitude as needed
        if self.nmoves == self.update_frequency {
            for mcmove in &mut self.moves {
                let direction = mcmove.1.direction();
                mcmove.0.update_acceptance(direction);
            }
            self.nmoves = 0;
        }
    }
}
```

src/core/src/sim/mc/moves/mod.rs
g-bauer: Not really, I use it in my code. It only limits the change though. We can easily run tests on what would be the most efficient number here (which would heavily depend on the system I guess).

With a large value, initially there will be very large changes which can lead to faster convergence to the target acceptance.

src/core/src/sim/mc/moves/mod.rs
Luthaf: It could be called before adding the move to the Monte-Carlo simulation. After adding it, you are right that there is no more control over the moves. See also how `MonteCarlo::add` send an error if a move is added after the simulation is started.

src/core/src/sim/mc/moves/rotate.rs
g-bauer: I see. You would then create a new move and set everything explicitly before adding it to the propagator?

Luthaf: Yes, this feels more clean to me.

src/core/src/sim/mc/moves/mod.rs
g-bauer: Agreed.

g-bauer:
All you wrote makes sense. Shifting counting into `propagate` will certainly save some troubles.

> for Hybrid MC (#16), the adjustable parameter is not the amplitude of the move, but the number of steps of the MD run (an integer and not a float).

That is true, future moves could have all kinds of adjustable parameters (number of configurational bias moves, number of MD steps,...). But aren't you shifting the problem towards `UpdateDirection` where `More` and `Less` will encapsulate floats?

How do we progress? I could implement your suggestions and commit the updates.

Luthaf:
> That is true, future moves could have all kinds of adjustable parameters (number of configurational bias moves, number of MD steps,...). But aren't you shifting the problem towards UpdateDirection where More and Less will encapsulate floats?

Yeah maybe. I think we need to pass the acceptance ratio, because some moves will need it. if other moves do not need the ratio, they always can ignore the value. But the fact that I needed to write `debug_assert!(ratio < 1.0); assert!(ratio > 0.0);` in the match do not please me. I'll think about a solution here.

> How do we progress? I could implement your suggestions and commit the updates.

Go for it! We always can rebase this branch later to clean the git history. I'll also have some time on Tuesday evening if you need some help.

g-bauer:
Still WIP. Maybe you can have a look at the changes. I decided to compute the scaling factor for the updates as method of `MoveCounter`. Still not very good for the general case (i.e. decide the number of MD steps or configurational bias moves according to acceptance). That way, it is completely decoupled from the move implementation and so I put it in `monte_carlo.rs`.

Other option: hand over the acceptance ratio and the target acceptance ...

TODO: 
- Better documentation
- How to set `update_frequency` in `MonteCarlo`?
- Communication with `input`

src/core/src/sim/mc/monte_carlo.rs
```diff
@@ -64,10 +67,21 @@ impl MonteCarlo {
                 we can not add new moves."
             );
         }
-        self.moves.push(mcmove);
+        self.moves.push((mcmove, MoveCounter::new(None)));
         self.frequencies.push(frequency);
     }
 
+    pub fn add_with_acceptance(&mut self, mcmove: Box<MCMove>, frequency: f64, target: Option<f64>) {
```
Luthaf: I feel like the name of this function should start with `add_move`, but I can not find a nice one. I though about `add_move_with_acceptance`, `add_move_auto_update`, ... 

What do you think?

Luthaf: The `target` parameter should be a f64 here. there is already a function for passing `None`.

src/core/src/sim/mc/moves/mod.rs
```diff
@@ -27,20 +28,28 @@ pub trait MCMove {
...
     /// `prepare`.
     fn apply(&mut self, system: &mut System);
 
-    /// Restore the system to it's initial state, if it has been changed in
+    /// Restore the system to it's initial state if it has been changed in
     /// `prepare`.
     fn restore(&mut self, system: &mut System);
+    
+    /// Update the displacement
+    fn change_range(&mut self, scaling_factor: Option<f64>);
```
Luthaf: This could have a better name, how about something like `update_amplitude(&mut self, factor: Option<f64>)`. 

Also, what does the Option means here? If it is `None`, no changes should be performed?

src/core/src/sim/mc/moves/rotate.rs
```diff
@@ -22,8 +22,10 @@ pub struct Rotate {
     newpos: Vec<Vector3D>,
     /// Normal distribution, for generation of the axis
     axis_rng: Normal,
+    /// Maximum values for the range of the range distribution of the angle
+    theta: f64,
     /// Range distribution, for generation of the angle
-    angle_rng: Range<f64>,
+    theta_range: Range<f64>,
```
Luthaf: This could just be `range`

src/core/src/sim/mc/moves/rotate.rs
```diff
@@ -153,4 +163,4 @@ fn rotation_matrix(axis: &Vector3D, angle: f64) -> Matrix3 {
     );
 
     return rotation;
-}
+}
```
Luthaf: Please leave an empty new line at the end of the files.

src/core/src/sim/mc/moves/translate.rs
```diff
@@ -20,34 +20,33 @@ pub struct Translate {
     molid: usize,
     /// New positions of the atom in the translated molecule
     newpos: Vec<Vector3D>,
-    /// Translation vector
-    delta: Vector3D,
+    /// Maximum displacement value
+    delta: f64,
     /// Translation range for random number generation
     delta_range: Range<f64>,
```
Luthaf: This could just be `range` too

src/core/src/sim/mc/moves/translate.rs
```diff
@@ -20,34 +20,33 @@ pub struct Translate {
...
+    /// Create a new `Translate` move, with maximum displacement of `dr`.
+    /// Translating only molecules with `moltype` type.
     pub fn with_moltype(dr: f64, moltype: u64) -> Translate {
         Translate::create(dr, Some(moltype))
     }
 
     /// Factorizing the constructors
     fn create(dr: f64, moltype: Option<u64>) -> Translate {
         assert!(dr > 0.0, "dr must be positive in Translate move");
-        let dr = dr / f64::sqrt(3.0);
```
Luthaf: I think you need to keep this division by sqrt(3) to ensure that the vector norm will be lower than the given dr.

src/core/src/sim/mc/moves/translate.rs
```diff
@@ -72,15 +71,15 @@ impl MCMove for Translate {
             return false;
         }
 
-        self.delta = Vector3D::new(
+        let displace_vec = Vector3D::new(
```
Luthaf: naming: displacement? delta?

Luthaf:
> How to set update_frequency in MonteCarlo?

What is wrong with a simple `MonteCarlo::update_frequency(&mut self, frequency: usize)` for that?

> Communication with input 

This can be made in a separated PR. That was part of the idea for separating the crates, you do not always need to update all of them at the same time.

Luthaf:
You will also need to rebase on top of master (as there are conflicts) and squash related commits together. [Here](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) is a small guide about it, just use `rebase -i master` instead of `HEAD~3`. I also found a longer explanation [here](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history).

src/core/src/sim/mc/moves/mod.rs
g-bauer: Yes you are right.

g-bauer:
I need to further work on the documentation. Skimming through the API doc, there is some confusing stuff. Also, I should add some examples.

Luthaf:
So Travis is failing on the `lumol-core` tests, [here](https://github.com/g-bauer/lumol/blob/f5d8e527c883d7b9afda35ebebd7724b514bd93a/src/core/src/sim/mc/monte_carlo.rs#L268-L274). This is missing the new `update_amplitude` function.

You can run the tests locally by doing `cd src/core; cargo test`. This will be easier to do when https://github.com/rust-lang/cargo/pull/3221 lands.

Luthaf:
> I need to further work on the documentation. Skimming through the API doc, there is some confusing stuff. Also, I should add some examples.

You will also need to add a few unit tests for the codecov check to pass. For example, you can test that `compute_scaling_factor` does return `None` when called too early, and return the right value of the ratio for the given values of `self.naccepted`  and `self.nattempted`. And same for the other functions.

Luthaf:
The codecov `diff hit` check is good for me, missing path where not previously checked. Do you want to add more doc/examples or should I merge this?

g-bauer:
I'd be happy if you merge if you think the coverage is ok.

Luthaf:
Merged 🎉 !

PR #54, opened 2016-11-17
Implement tail corrections to the energy and virial

### LINKED ISSUE #28: Add computation for truncated potentials with tail corrections

Luthaf:
The core logic is implemented [here](https://github.com/lumol-org/lumol/blob/fc2a0895d76f2b806673b152062233883534c3ad/src/core/src/sys/energy.rs#L55-L70) for energy and [here](https://github.com/lumol-org/lumol/blob/fc2a0895d76f2b806673b152062233883534c3ad/src/core/src/sys/compute.rs#L196-L207) for the virial, using the [`PairInteraction::tail_energy` and `PairInteraction::tail_virial`](https://github.com/lumol-org/lumol/blob/fc2a0895d76f2b806673b152062233883534c3ad/src/core/src/energy/pairs.rs#L214-L229) functions.

This need a bit more tests, @g-bauer what was your strategy for checking against NIST data?

Fix #28


g-bauer:
> This need a bit more tests, @g-bauer what was your strategy for checking against NIST data?

I use the [NIST reference calculations](https://www.nist.gov/mml/csd/chemical-informatics-research-group/lennard-jones-fluid-reference-calculations). I build an example, where the configurations offered by NIST are read and the system energy is calculated and checked vs. the values from NIST. I used the examples to test the energies without tail corrections. I'll send you the examples.

Regarding the PR, I think the name `tail_virial` is a bit misleading since you only use it to compute a correction to the pressure.

Luthaf:
> Regarding the PR, I think the name tail_virial is a bit misleading since you only use it to compute a correction to the pressure. 

It is used in the computation of the `Virial` tensor, which in turn is used for the pressure.

I agree that the name `Virial` for this tensor is bad. It is more something like the virial contribution to the stress tensor. I'd like to improve the name, but I have no idea of a better one for the moment.

Luthaf:
I added the test comparing against NIST, but a kind of potential is missing to compare with the LFS column.

Linear Force-shift potential can be added later (and are actually an easy thing to do), see issue #55 for this.

PR #58, opened 2016-11-28
Add resize MC move.

g-bauer:
This adds a MC move for changing the size of the simulation cell.

Todo issue #59 :
I added a check to compare the smallest inscribed spheres' radius vs the cut off radius `rc_max`. When the cell gets too small (so that the cut off is too large) it should cancel the simulation and report an error. I'd like to get the (maximum) cut off out of the `PairInteractions` when invoking `new()` but it is not clear how to get it at the moment.

@Luthaf we wrote the main part of this together. I created a new branch since the history was messed up in the old branch. Now, you are not reported as author. Can we change this somehow?


Luthaf:
> @Luthaf we wrote the main part of this together. I created a new branch since the history was messed up in the old branch. Now, you are not reported as author. Can we change this somehow?

This should be possible, I've never done it. Maybe something using rebase, and then amend the commit to add an author. I don't even know if you can have more than one author for a commit. Anyway, you can leave it as it is, I can live without being cited as author on this commit 😉.

I'll give this PR an eye later today.

g-bauer:
> I can live without being cited as author on this commit 😉

So do I. If we find a way to add/change author, I'd be happy.

> I'll give this PR an eye later today.

Great! I will write a test case.

Luthaf:
> So do I. If we find a way to add/change author, I'd be happy.

I checked that, and git can not associate two authors with the a given commit. You can add me in the commit message (`Implemented with ...); or I can do this before merging.

src/core/src/sim/mc/moves/resize.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// displacement of `delta`.
+    pub fn new(pressure: f64, delta: f64) -> Resize {
+        assert!(delta > 0.0, "delta must be positive in Resize move");
+        // TODO: set rc_max from the largest cut off in PairPotentials
+        Resize {
+            delta: 0.,
+            range: Range::new(-delta, delta),
+            old_system: System::new(),
+            pressure: pressure,
+            rc_max: 100.0, // arbitrarily large value
```
Luthaf: You could use `std::f64::INFINITY` here.

src/core/src/sim/mc/moves/resize.rs
```diff
@@ -0,0 +1,112 @@
...
+        let volume = system.volume();
+        let scaling_factor = f64::cbrt((volume + self.delta) / volume);
+        // change the simulation cell
+        system.cell_mut().scale_mut(Matrix3::one() * scaling_factor);
+        let new_cell = system.cell().clone();
+        // check the radius of the smallest inscribed sphere and compare to the 
+        // cut off distance.
+        // abort simulation when box gets too small
+        if new_cell.lengths_as_vec().iter().any(|&d| 0.5 * d < self.rc_max) {
+            fatal_error!(
```
Luthaf: I'd argue this should not be a fatal error. It should send an error message saying that the energy will be wrong for now on, but not abort the process, as `fatal_error!` does.

src/core/src/sim/mc/moves/resize.rs
```diff
@@ -0,0 +1,112 @@
...
+            particle.position = new_cell.cartesian(&fractional);
+        }
+        true
+    }
+
+    fn cost(&self, system: &System, beta: f64, cache: &mut EnergyCache) -> f64 {
+        cache.unused();
+        // get system energy before change: this is stored in `self.old_system`
+        let old_energy = self.old_system.potential_energy();
+        // get
```
Luthaf: Unfinished comment

src/core/src/sys/cells.rs
```diff
@@ -174,6 +174,23 @@ impl UnitCell {
         return (x, y, z);
     }
 
+    /// Get a vector of the distances between faces of the unit cell
+    pub fn lengths_as_vec(&self) -> Vec<f64> {
```
Luthaf: We do not need another function here, the `UnitCell::lengths` can do this too! You seems to need iteration on the lengths, so why not change the signature to be `pub fn lengths(&self) -> [f64; 3]`? This way you get iteration, and do not pay for a dynamic allocation.

The current function is only [used once](https://github.com/lumol-org/lumol/blob/677d697b91ba41f0ef6c578968cf1ad8d551df21/src/core/src/energy/global/ewald.rs#L124-L134), for the exact same reason as here.

src/core/src/sim/mc/moves/resize.rs
g-bauer: Well, wrong energies lead to wrong configurations. In the end, the ensemble averages computed are incorrect. We could adjust the cutoff so that it fits into the new box. That is done in some codes. Personally, I think this solution is also not a very good idea, since it essentially changes the force field.

src/core/src/sys/cells.rs
g-bauer: Good catch! I will change that. Actually, I tried to get the maximal value out of the tuple but it was kind of messy. I am not used to the fact that there is no `max` function for arrays and tuples.

src/core/src/sim/mc/moves/resize.rs
g-bauer: Thanks for the catch.

src/core/src/sim/mc/moves/resize.rs
g-bauer: I recognized that there has to be a zero. `std::f64::INFINITY` would lead to the error, since any side length is smaller than infinity.

src/core/src/sim/mc/moves/resize.rs
Luthaf: I don't like the idea of changing the cutoff too.

I prefer to leave control to the user, instead of doing things behind him. We can leave a fatal_error, and reconsider if this is being very annoying.

g-bauer:
> You can add me in the commit message (`Implemented with ...); or I can do this before merging.

Please, be so kind and do this when merging :smile:

Luthaf:
This is missing some [integration tests](http://doc.crates.io/manifest.html#integration-tests) before merging.

I would at least have a check that for a Lennard-Jones fluid, the pressure does converge to the requested value, adding a new case [here](https://github.com/lumol-org/lumol/blob/677d697b91ba41f0ef6c578968cf1ad8d551df21/tests/mc-helium.rs).

> Please, be so kind and do this when merging 😄

Yes, will do so!

g-bauer:
I will add test case(s) tonight.

Luthaf:
> A practical question: in #58, I added an example case. Somehow, I couldn't manage to make this a #[test]. I compiles, but is not able to find the configuration file. I placed the config in lumol/tests/data/ and the test in lumol/tests. Using cargo test --test my_test --release fails.

Are you reading the file like this ?

```
let data_dir = Path::new(file!()).parent().unwrap();
let configuration = data_dir.join("data").join("argon.xyz");
let mut system = Trajectory::open(configuration)
                                            .and_then(|mut traj| traj.read())
                                            .unwrap();
```

Because your error looks like a problem with the current directory of the test process.

g-bauer:
That's exactly what I am doing. I copied this part from the existing test `helium.rs` (which throws the same error).

Luthaf:
Weird. I'll clone this branch and have a look. Can you commit your changes and push them ?

g-bauer:
Changes are commited. Both, example and test compile. Example just works when I call it from inside `lumol/examples/`. Test does not run.

I did:
```
$ cd /lumol/tests
$ cargo test --test mc_npt_argon
```

Luthaf:
Ok, I got it ! You need to run cargo from the `lumol` folder, not the `lumol/tests` folder. 

`Path::new(file!()).parent().unwrap();` will give you the directory containing the current file, so `configuration` will be `tests/data/argon.xyz`. 

At the same time, the examples do not use the `Path::new(file!()).parent().unwrap();` trick, so they must be run from the example directory. I realize this is actually confusing, we might want to change examples to use paths from the root directory too. Or we could change everything to use absolute paths only.

g-bauer:
> I would at least have a check that for a Lennard-Jones fluid, the pressure does converge to the requested value, adding a new case here.

Energies and volumes look good. I need to perform some long runs over night to actually check the statistics. Internal pressure seems (no good statistics) to work fine for the case without tail corrections. With tail corrections, there is an issue. I'll check that.

Luthaf:
Can you rebase this on master now? All the fixes should be here !

g-bauer:
I did a lot more work on this. I will rebase and push the changes later today.

g-bauer:
### This is still WIP

I changed a lot in the latest iteration. Ran some short simulations, while energy and density seem to be correct, internal pressure is not correct yet.

 - I changed the struct to contain the `new_system` instead of the old/current system. This is more in line with the translation and rotation moves, only the `apply` function will modify the system.
- Also, positions of molecules ares scaled such that the intramolecular distances are unchanged.
- At the moment, energy computation is inefficient. We can fetch the current (old) energy from cache. The new energy is computed from scratch. Since there is noch computation via cache yet, when applying the move, we recompute the whole energy again (`cache.unused() -> cache.init`).

src/core/src/sim/mc/moves/resize.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn describe(&self) -> &str {
+        "resizing of the cell"
+    }
+
+    fn setup(&mut self, system: &System) {
+        // get the largest cutoff of all intermolecular interactions in the
+        // system.
+        // TODO: include elecetrostatic interactions
+        self.rc_max = system.interactions()
+            .all_pairs()
```
Luthaf: This would look better if aligned with the `.interactions()` =)

Luthaf:
> Also, positions of molecules ares scaled such that the intramolecular distances are unchanged.

Nice !

> At the moment, energy computation is inefficient. We can fetch the current (old) energy from cache. The new energy is computed from scratch. Since there is noch computation via cache yet, when applying the move, we recompute the whole energy again (cache.unused() -> cache.init).

We can add a `EnergyCache::recompute_all(system) -> f64` (bad name) function that recompute all the terms, and call it [here](https://github.com/lumol-org/lumol/pull/58/files#diff-3ef79bc895ec988f9ccf59155455db57R111). The `EnergyCache::recompute_all` function can set the right `updater` closure to skip computing everything twice in the case we accept the move.

g-bauer:
Another take would be to make this more like the translation move, i.e. not to clone the whole system, but to get only the components that actually change. For example

```rust
/// Monte-Carlo move that changes the size of the simulation cell
pub struct Resize {
    /// Delta for translation of the box length
    delta: f64,
    /// Sampling range for volume scaling
    range: Range<f64>,
    /// Positions after applying changes to the simulation cell
    new_pos: Vec<Vector3D>,
    /// New simulation cell
    new_cell: UnitCell,
    /// target pressure
    pressure: f64,
    /// largest cut off diameter of `PairPotentials`
    rc_max: f64,
}
``` 

We could then create a caching function like
```rust
pub fn resize_cell_cost(&mut self, system: &System, new_pos: &[Vector3D], new_cell: UnitCell) -> f64
```
that only recomputes intermolecular changes. This would work nicely with the  `update` closure. 
There are currently two things preventing such an implementation: 1) We'd need to also implement this for electrostatic interactions, 2) `apply`ing the move is not as elegant; we'd have to reiterate over all particles in the system and change each position individually.

g-bauer:
I added a function `move_rigid_molecules_cost` (maybe change to `move_resize_cell_cost`?) that takes a (new) `system` and recomputes only the intermolecular interactions + tail corrections + global interactions from scratch which makes `resize` more efficient to use. I didn't dive into the caching of `coulomb` and `global` - these are recomputed from scratch at the moment, I added a `TODO`.

I also changed the sign of the tail correction to the virial. I did two quick runs for Lennard Jones particles and compared to NIST. Energies and densities are in very good (as far as I can tell from the short runs) agreement, internal pressure fits external pressure.

I will add tests/examples for TraPPE Ethane and Water to check if molecules/electrostatics also work as intended.

src/core/src/sys/cache.rs
```diff
@@ -454,4 +457,38 @@ mod tests {
...
+        let system = testing_system();
+        let mut cache = EnergyCache::new();
+        let old_e = system.potential_energy();
+        cache.init(&system);
+
+        let delta = Vector3D::new(1.0, 0.5, -0.5);
+
+        let mut new_system = system.clone();
+        // translate the center of mass
+        for pi in &system.molecules()[0] {
```
Luthaf: This can be `system.molecule(0)` until we get better API for molecules.

src/core/src/sys/compute.rs
```diff
@@ -201,7 +200,7 @@ impl Compute for Virial {
                 let nj = composition[&j] as f64;
                 let potentials = system.interactions().pairs(i, j);
                 for potential in potentials {
-                    virial -= 2.0 * PI * ni * nj * potential.tail_virial() / volume;
+                    virial += 2.0 * PI * ni * nj * potential.tail_virial() / volume;
```
Luthaf: This should be in a separated commit (commits should do only one thing), and maybe with an explanation message (the commit message) and associated tests.

It is fine for this one, can you try to do it for your next commits?

Luthaf:
> Energies and densities are in very good (as far as I can tell from the short runs) agreement, internal pressure fits external pressure.

Yay ! \o/

> I didn't dive into the caching of coulomb and global - these are recomputed from scratch at the moment, I added a TODO.

This can be a separated PR.

src/core/src/sys/compute.rs
g-bauer: You are right. I'll try to keep separate issues in separate commits in the future.

g-bauer:
I added functionalities to wrap molecules into the simulation cell. I use these to shift molecules as a whole, as soon as the center of mass of the molecule leaves the cell (some particles may still be outside the cell, but this way the com is always inside the cell). I used this mainly to debug some weird stuff happening to the bond lengths. In the end, the molecule wrapping is not necessary but it might come in handy in the future, for example when writing the configuration to a file.

I added a property `dr_max` to `Translate` which is set via `Translate::setup` that is used to limit the amplitude for translations. I found that for the ethane system, i.e. a dilute gas phase, the amplitude explodes since virtually all translations are accepted. That leads to numerical problems.

I added an example case for ethane. I found very good agreement for energy and density w.r.t other codes, and internal pressure agrees very well with external pressure. Going to add a case for SPC/E water.

examples/mc_npt_ethane.rs
```diff
@@ -0,0 +1,107 @@
...
+fn get_system() -> System {
+    let data_dir = Path::new(file!()).parent().unwrap().join("data");
+    let configuration = data_dir.join("ethane.xyz");
+    let mut system = Trajectory::open(configuration)
+                                .and_then(|mut traj| traj.read_guess_bonds())
+                                .unwrap();
+    system.set_cell(UnitCell::cubic(100.0));
+
+    // Add intermolecular interactions
+    // TraPPE parameters for CH3 - note that masses are not correct.
```
Luthaf: You can iterate over the particles and set the right mass like this:

```rust
for particle in &mut system {
    assert_eq!(particle.name(), "C");
    particle.mass = ...;
}
```

examples/mc_npt_ethane.rs
```diff
@@ -0,0 +1,107 @@
...
+    system.interactions_mut().add_pair("C", "C", pairs);
+
+    // Add bonds: we use fixed bond lengths assuming
+    // the equilibrium bond length. This means that both
+    // energy as well as virial for the bond potential are
+    // zero. Hence, we use a `NullPotential`.
+    let bond = Box::new(NullPotential{});
+    system.interactions_mut().add_bond("C", "C", bond);
+
+    // Check system if bonds are guessed correctly.
```
Luthaf: // check if bonds ...

examples/mc_npt_ethane.rs
```diff
@@ -0,0 +1,107 @@
...
+    // accepted no matter what the amplitude will be.
+    mc.add_move_with_acceptance(
+        Box::new(Translate::new(delta_trans)), 50.0, 0.5);
+    mc.add_move_with_acceptance(
+        Box::new(Rotate::new(delta_rot)), 50.0, 0.5);        
+    mc.add_move_with_acceptance(
+        Box::new(Resize::new(pressure, delta_vol)), 2.0, 0.5);    
+    mc.set_amplitude_update_frequency(500);
+
+    // Setup simulation.
```
Luthaf: btw, why not use an input file here ? It does not seems to be too complicated.

src/core/src/sys/cells.rs
```diff
@@ -276,6 +276,30 @@ impl UnitCell {
         }
     }
 
+    /// Return a new vector that lies inside the simulation cell.
+    ///
+    /// If `vect` is already in the cell or the `CellShape` is
+    /// `Infinite`, the resulting vector is identical to `vect`.
+    pub fn get_wrapped_vector(&self, vect: &Vector3D) -> Vector3D {
```
Luthaf: Why is this needed? Couldn't the `wrap_vector` function be used instead? i.e. replace

```rust
let wrapped = cell.get_wrapped_vector(original);
```

by 

```rust
let mut wrapped = original.clone();
cell.wrap_vector(&mut wrapped);
```

src/core/src/sys/compute.rs
```diff
@@ -236,8 +236,8 @@ impl Compute for PressureAtTemperature {
         let virial_tensor = system.virial();
         let virial = virial_tensor.trace();
         let volume = system.volume();
-        let natoms = system.size() as f64;
-        return natoms * K_BOLTZMANN * self.temperature / volume + virial / (3.0 * volume);
+        let nmols = system.molecules().len() as f64;
```
Luthaf: I do not get why this is needed. Can you explain?

examples/mc_npt_ethane.rs
g-bauer: Ah I see. Here, it would not make a difference, since we don't output the density.

examples/mc_npt_ethane.rs
g-bauer: Good catch, thanks!

examples/mc_npt_ethane.rs
g-bauer: I wrote an input file - I can also add it to the PR. Since we can't use the input to set up the whole simulation (tail correction, update frequency, resize move) yet, I though it is more rigorous to implementation everything "by hand". I agree though, we should add the same example using an input file once we have most of the above mentioned things covered.

src/core/src/sys/cells.rs
g-bauer: Well, this solution did not came into my mind 😄 . I will get rid of it, thanks!

src/core/src/sys/compute.rs
g-bauer: You mean why use the number of molecules instead of number of atoms? I can try to explain the idea in some sentences - if you want, I can also send you a derivation via mail.

The idea is, that for a system consisting of noninteracting molecules (no intermolecular interactions but generic intramolecular potentials) we can write the NVT partition function as a product of single-molecule partition functions. 

We can further do a coordinate transformation such that the positions of the atoms of a molecule (let's say that there are `n` atoms per molecule) are reformulated to yield the center of mass position of the molecule plus the `n-1` *relative positions* in the molecule. The intra molecular energy of a single molecule does not depend on the center of mass position of the molecule, so we can perform the integration over the center of mass position in the partition function directly. This integral yields `V` (the volume) for every molecule, i.e. for `N` molecules this is `V^N`.

The rest of the partition function is no longer a function of the volume, but of temperature and the number of atoms in the molecule. The pressure is computed by taking the derivative of the Helmholtz energy w.r.t the volume. The only volume dependent term of the ideal gas partition function of a molecular system is `V^N`. Since we have to take the logarithm of the partition function, we get `p = kT * d/dV (N log V) = NkT/V`.

Does that make sense? It's a bit difficult to describe in words.

examples/mc_npt_ethane.rs
Luthaf: We can change this to use input file later, sure.

src/core/src/sys/compute.rs
Luthaf: I'll try to write this down tomorrow, and double check in the other codes I know =)

Does this mean that computing the pressure by using the number of atoms will yield the wrong result?

g-bauer: > Does this mean that computing the pressure by using the number of atoms will yield the wrong result?

Yes, I think so, but let me double check tomorrow as well. :) FWIW I got the right pressure (i.e. external == internal) performing the simulation with the number of molecules instead of atoms.

Luthaf: OK, I think I got it. The real formula for computing pressure does not include the number of atoms nor the number of molecules, it is the number of degree of freedom over the dimensionality of the system. Which roughly equals the number of atoms for flexible molecules, and the number of molecules for rigid molecules.

[LAMMPS](http://lammps.sandia.gov/doc/compute_pressure.html) documentation explains this quite well, but does not give any algorithm to compute it.

[DL_POLY]() manual page 68 has a "constraint" term in the pressure equation, because in MD rigid molecules only comes from constraint algorithms.

[Tohwee](http://towhee.sourceforge.net/algorithm/pressure.html) uses 4 different algorithms to compute the pressure in MC simulations, I do not know which one would be the best here.

In the [Allen & Tildesley book](https://www.amazon.com/Computer-Simulation-Liquids-Science-Publications/dp/0198556454) page 47, they follow your definition using the number of molecules, whereas in the [Frenkel & Smit](https://www.amazon.com/Understanding-Molecular-Simulation-Second-Computational/dp/0122673514) book page 84, they use the number of atoms.

I do not think that implementing the correct version of the code (using degree of freedom) is doable here, so maybe we can revert this change and indicate clearly in the documentation that this does not work for rigid molecules? We should also open a separated issue to track work on this problem.

g-bauer: Mhm I did the derivation for non-rigid molecules. It is also completely derived in Tuckerman's Statistical Mechanics (in the solutions to the exercises, I can send you those). From what I can see, Towhee uses the number of molecules exclusively. Is Towhee restricted to rigid molecules?

I don't have Frenkel & Smit here, I will have a look on monday.

g-bauer:
Concerning the current state of this PR:

- How do we decide on the ideal gas pressure discussion? (`let nmols = system.molecules().len() as f64;`)?
- I added a test for charged systems (SPC/E) which shows weird behavior concerning the internal pressure while energies and densities are reasonable. Pressure issues might resolve when the bug fix from PR #75 lands.
- Documentation as well as input implementation are done in a separate branch (finished but not polished)

Do you have additional comments or suggestions?

Luthaf:
> How do we decide on the ideal gas pressure discussion? (let nmols = system.molecules().len() as f64;)?

In the Tuckerman book, the derivation is done for *"N identical **noninteracting** molecules"*, and I could not manage to replicate this in the case with arbitrary interactions between molecules. Moreover, in the equation (5.9.3) of the same book, Tuckerman uses the number of degrees of freedom to compute the instantaneous kinetic energy (which is at the origin of the term we are discussing here) instead of the number of particles. I still need to find the time to make the full derivation by hand, but I am still convinced the right number is the number of degrees of freedom, not the number of particles or of molecules.

---

While we are working on the science, we can still merge this code. We can do the following:

1. Leave the code as it is;
2. implement a new `Compute` in the butane test file that uses the number of molecule;
3. add a fixme to both pressure compute algorithms (in lumol-core and in the butane test) and open an issue about it;
4. Fix the issue later.

If you are OK with that, we can merge this code and get everything else working later =)

g-bauer:
> In the Tuckerman book, the derivation is done for "N identical noninteracting molecules", and I could not manage to replicate this in the case with arbitrary interactions between molecules.

But why do you need that in this case? To compute the ideal gas part we set the intermolecular interactions to zero, right?

Never mind, you are right, we should work this out in detail in a different issue/PR 😄 

> If you are OK with that, we can merge this code and get everything else working later =)

Sure! I think the move itself works. The rest should have its own issue. Should I change the computation back to the number of atoms in pressure/stress?

Luthaf:
>  Should I change the computation back to the number of atoms in pressure/stress? 

Please do, and I'll change the authoring information for the first commit, and then merge =)

g-bauer:
I changed it back.

examples/mc_npt_argon.rs
```diff
@@ -0,0 +1,93 @@
...
+    // T  = 102.231042155 K
+    // V  = 25306.1891827 A^3
+    // p  = 20.1586274874 bar
+    // L  = 29.3590670058 A
+
+    // We use this state to compare to NIST NVT data
+    
+    let temperature = units::from(102.231042155, "K").unwrap();
+    let pressure = units::from(20.1586274874, "bar").unwrap();
+    //let mut mc = MonteCarlo::from_rng(temperature, rng); // This is not working, why?
```
Luthaf: Cleanup? And maybe open an issue with the error if this is not working =)

examples/mc_npt_spce.rs
```diff
@@ -0,0 +1,125 @@
...
+    system.interactions_mut().add_angle("H", "O", "H", angle);
+
+    // Set charges
+    let h = 0.42380;
+    let o = - 2.0 * h;
+    for p in &mut system {
+        match p.name() {
+            "H" => p.charge = h,
+            "O" => p.charge = o,
+            _ => panic!("Particle {}", p.name()),
```
Luthaf: Maybe use a better message here ("Unknown particle in charge setting")? And replace `p` by `particle`?

src/core/src/sys/systems.rs
```diff
@@ -282,6 +282,12 @@ impl System {
     #[inline] pub fn size(&self) -> usize {self.particles.len()}
 
     /// Return the center-of-mass of a molecule
+    ///
+    /// # Warning
+    ///
+    /// This function does not check for the particles' positions'
+    /// nearest images. To use this function properly, make sure
+    /// that all particles of the molecule are adjacent.
```
Luthaf: I don't understand this warning. Is it in the case where we have some particles of the molecule on a side of the cell and some other particles on the other side? Why not check particles nearest image here?

src/core/src/sys/systems.rs
```diff
@@ -309,6 +321,33 @@ impl System {
...
+    /// # Note
+    ///
+    /// If the `CellShape` is `Infinite` there are no changes
+    /// to the positions.
+    pub fn wrap_molecule(&mut self, molid: usize) {
+        // This catch is not necessary, since for `Infinite`
+        // cells, a wrap will return just the same vector and
+        // hence `delta` will be zero; but it is faster.
+        match self.cell().shape() {
+            CellShape::Infinite => (),
```
Luthaf: This is not needed, as the `wrap_vector` function already manage the Infinite cell case. The same code ca be used everywhere.

src/core/src/sys/systems.rs
```diff
@@ -638,6 +677,15 @@ impl IndexMut<usize> for System {
     }
 }
 
+impl Index<Range<usize>> for System {
+    type Output = [Particle];
+    #[inline]
+    fn index(&self, index: Range<usize>) -> &[Particle] {
+        // Index::index(&self.particles, index)
```
Luthaf: cleanup?

tests/mc_npt_argon.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    // set thermodynamic state (T,p) 
+    // in LJ units, this is:
+    // T* = 1.0
+    // p* = 0.03
+    // We use this state to compare to [1]
+    
+    let temperature = units::from(120.0, "K").unwrap();
+    let pressure = units::from(12.6, "bar").unwrap();
+    //let mut mc = MonteCarlo::from_rng(temperature, rng); // This is not working, why?
```
Luthaf: cleanup?

tests/mc_npt_argon.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+#[test]
+fn mc_npt_argon() {
+    START.call_once(|| {Logger::stdout();});
+    let mut system = get_system();
+    
+    // rng: set this here in case we want to change the seed
+    //let mut rng = Box::new(rand::XorShiftRng::new_unseeded());
+    //rng.reseed([2015u32, 42u32, 3u32, 12u32]);
```
Luthaf: cleanup? This is not adding a lot of information to this test.

tests/mc_npt_argon.rs
```diff
@@ -0,0 +1,87 @@
...
+        Box::new(Resize::new(pressure, delta_vol)), 2.0, 0.3);
+    // if a move was called 500 times, perform an update
+    mc.set_amplitude_update_frequency(500);
+        
+    let mut simulation = Simulation::new(Box::new(mc));
+    
+    // run simulation for at least 500 cycles (so that a volume update is performed)
+    simulation.run(&mut system, 500000);
+
+    // [1] Lotfi et al.:
```
Luthaf: Why are we not checking that in `assert_approx_eq!`

Luthaf:
I rebased the whole branch on top of master, and have added our names to the first commit.

g-bauer:
Thanks for the review. Can I make my changes locally and then simply push as usual or do I have to fetch your changes beforehand?

Luthaf:
You have to get the changes first. Use `git fetch origin` and then `git reset --hard origin/resize-cell` (assuming `github.com/g-bauer/lumol` is your `origin`)

tests/mc_npt_argon.rs
g-bauer: What do you mean? Checking versus energy and density from the paper?

src/core/src/sys/systems.rs
g-bauer: Should we keep indexing function via `Range`? We currently don't use it.

tests/mc_npt_argon.rs
Luthaf: Yes, or move this from the test folder to the examples folder. It does not seems to test anything yet =)

src/core/src/sys/systems.rs
Luthaf: If we do not use it, let's remove it for now. It is easy to add back if needed.

src/core/src/sys/systems.rs
g-bauer: > Is it in the case where we have some particles of the molecule on a side of the cell and some other particles on the other side?

Exactly. To avoid that, we can use `wrap_molecule`. It get's messy when an input file is used and parts of a molecule are already wrapped.

tests/mc_npt_argon.rs
g-bauer: We already have a NPT example for argon. I remove this one. The simulation has to run very long to get comparable results.

g-bauer:
Changed according to review. @Luthaf Again, thanks for your patience and your time investment. 

The examples for SPCE water and ethane are still in the example folder although they currently do not produce correct samples. For ethane, pressure is wrong due to using the number of atoms for the ideal gas pressure which we discussed above and which will be resolved in a separate PR. The same is true for SPCE which is also affected by the bugs that were found and worked on in #75.

I left the current implementation of the center of mass computation - without checking if a molecule is partly wrapped. This would result in a wrong center of mass. I'd also address changes in a separate PR.

src/core/src/sys/systems.rs
g-bauer: Yep, I removed it.

src/core/src/sys/systems.rs
g-bauer: That's true. Removed the match statement.

src/core/src/sys/systems.rs
g-bauer: I would like to address this in a separate PR, if that's ok for you.

examples/mc_npt_spce.rs
g-bauer: Thanks for the comment, I added a better description.

src/core/src/sys/systems.rs
Luthaf: That's OK.

Luthaf:
Thank you @g-bauer, that was an epic PR! More than 6 month between the initial implementation and the merge, that is the best I have done.

g-bauer:
> Thank you @g-bauer, that was an epic PR! More than 6 month between the initial implementation and the merge, that is the best I have done.

Yeah! 🎉 Thanks @Luthaf for mentoring!

PR #61, opened 2016-11-30
Use f64 standard functions for angle conversion

odarbelaeze:
@g-bauer noticed this and it was a quick refactor.

Now we use [to_degrees](https://doc.rust-lang.org/1.6.0/std/primitive.f64.html#method.to_degrees) and [to_radians](https://doc.rust-lang.org/1.6.0/std/primitive.f64.html#method.to_radians) to do that transformation instead of an inline function.


Luthaf:
Thanks!

Issue #18, opened 2016-03-02
Where should we put our particles


Luthaf:
Currently, the particles are stored contiguously in a vector, and the molecules have an `usize` field pointing to the first particle of the molecule in this vector.

``` rust
struct System {
    particles: Vec<Particle>
    molecules: Vec<Molecule>,
}

struct Molecule {
    first: usize
}
```

Pro of this design
- good cache locality for particles
- easy to iterate over all particles in the system. A double loop over pairs is:

``` rust
for (i, pi) in system.iter().enumerate() {
    for pj in system.iter().skip(i + 1) {
        // Do work
    }
}
```

Cons of this design
- hard to iterate over particles in a molecule, and corresponding double loops are confusing.

``` rust
for molecule in system.molecules() {
    for i in &molecule {
        for j in &molecule {
            if j <= i {continue;}
            let pi = system[i];
            let pj = system[j];
            // Do work
        }
    }
}
```

Making the particles owned by molecules is not a solution, because it make it very hard to iterate over pairs in the whole system, and add an extra indirection layer.

A possible solution, with a POC implementation [here](https://gist.github.com/Luthaf/12fc5e0cbbeae3cf31e4) would be to store a pointer to the first atom in the molecule, together with the molecule size. This would make it easier to work with molecular code, but add some `unsafe` usage. 

In particular, the pointer of all molecules must be updated when adding a new particle to the system, because the vector may allocate additional memory.

So is it worth it to add complexity in the library code in order to simplify the user code?

Luthaf:
Another solution would be to have two separated types: an "owning" molecule and a reference molecule.

The owning molecule (called `Molecule`) would contain a `Vec<Particle>`, and can be manipulated without a `System`. A reference molecule (`MoleculeRef<'a>`) would contain a slice of particles `&'a [Particle]` borrowed from a system, and thus can not be manipulated without a `System`. The owned molecule would then implement `Deref<Target=MoleculeRef>`, and most of the utility functions for molecules would be implemented on `MoleculeRef`.

This allow users to manipulate particles through the molecules, without loosing cache locality.

g-bauer:
I am not sure if I got that. Could you elaborate about cache locality? I can see however that the current implementation is a bit uncomfortable to use. Are there problems with something like this (using horrible naming):

```rust
// method of System

/// Returns the particles of the molecule with index `id`
#[inline] 
pub fn particles_in_molecule(&self, id: usize) -> &[Particle] {
[...]

PR #65, opened 2016-12-07
Add function to compute a molecules' center-of-mass

g-bauer:


Luthaf:
Looks good to me, we will need to update this when a solution to #18 is found.

Could you also use this function in places where the COM is computed? I believe there is some usage of it in the MC moves, and maybe elsewhere too.

g-bauer:
I know it is used in `Rotate` and I will use it in `Resize`. I will look through the code and add it accordingly.

Luthaf:
And about the coverage limit, it changed in #67, so if you can rebase on top of master, you will get better numbers.

g-bauer:
@Luthaf Do you have more suggestions or can we merge this?

Luthaf:
That was all of it!

PR #66, opened 2016-12-07
Better testing for input files

### LINKED ISSUE #42: Better testing for bad inputs
Currently, bad input are tested by checking that the result of the input read is an error. But we do not control the source of the error. Allowing to specify the expected error inside the bad input file would guarantee that we do not get another error that the one wanted (bad TOML, IO error, ...).

Bad input files could use comments to specify the expected input error:

``` toml
[input]
versn = 1
#^ Missing 'version' key in 'input' table
```

Luthaf:
This PR changes the tests of the `input` crate to test for the actual error message send in case of bad input files. Not doing it meant that some files contained a different error than the expected one (for example [this one](https://github.com/lumol-org/lumol/commit/7025b1848e23531b1eced62af0ed393b69b61ad1#diff-18045255207d79898ae4ead3a81b1f72R4) where the error was about a missing `systems` key).

In order to do so, 2194d1c2bc44ef6b5671d7946adb1a7e5570d9ad move all tests in the `input/tests` folder, and uses [rustc-test](https://crates.io/crates/rustc-test) to [generate tests dynamically](https://github.com/lumol-org/lumol/commit/2194d1c2bc44ef6b5671d7946adb1a7e5570d9ad#diff-193a46787179d6265193c6a692224843R20).

Then 7025b1848e23531b1eced62af0ed393b69b61ad1 actually [check for the error message](https://github.com/lumol-org/lumol/commit/7025b1848e23531b1eced62af0ed393b69b61ad1#diff-193a46787179d6265193c6a692224843R32) and fix the bad input files to make them correspond to the expected error message. It also improves some error messages on the way.

I'd like to know if the test code is clear enough for you, or if I should try to make it clearer, so that other person can add more tests when touching to the `input` crate.

Fix #42


Luthaf:
@g-bauer, any comments here ? If you do not have time to review this is fine, just tell me and I'll merge.

I am pretty confident the code is OK, I would mainly like to know if it is understandable.

g-bauer:
It is very clear: you compare the error messages (starting with `^#`) from the test files with the actual coded error messages. 

I only skimmed over the actual messages but they look clear and understandable. In any case, it is very clear how to improve them if needed. I like it, nice work.

Luthaf:
Thank you for the review! 

I'll merge this PR, but I was thinking it would be better if no one merged his own pull requests, but rather the one doing the review would merge the PR. What do you think?

g-bauer:
>  it would be better if no one merged his own pull requests, but rather the one doing the review would merge the PR. What do you think?

That is fine with me. Although I might need a little help in the beginning.
Am I able to do this (technically - do I have the rights to do that)? I recognized that I can't add labels to issues (also when I create new ones).

Luthaf:
> Am I able to do this (technically - do I have the rights to do that)? 

I though you could, and then I checked again, and the default permission was set to read, not write. 

Now you should be able to!

PR #69, opened 2016-12-13
Refactor molecules

Luthaf:
Simple refactoring of the molecule type and associated structs.

### DIFF
--- a/src/core/src/sys/connect.rs
+++ b/src/core/src/sys/connect.rs
@@ -1,0 +1,163 @@
+// Lumol, an extensible molecular simulation engine
+// Copyright (C) 2015-2016 G. Fraux — BSD license
+
+//! Data about bonds and angles in the system.
+use std::cmp::{min, max};
+
+/// A `Bond` between the particles at indexes `i` and `j`
+///
+/// This structure ensure an unique representation of a `Bond` by enforcing
+/// `i < j`
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Bond {
+    i: usize,
+    j: usize,
+}
+
+impl Bond {
+    /// Create a new Bond between the particles at indexes `first` and `second`
+    pub fn new(first: usize, second: usize) -> Bond {
+        assert!(first != second);
+        let i = min(first, second);
+        let j = max(first, second);
+        Bond{i: i, j: j}
+    }
+
+    /// Get the first particle in the bond
+    #[inline] pub fn i(&self) -> usize {self.i}
+
+    /// Get the second particle in the bond
+    #[inline] pub fn j(&self) -> usize {self.j}
+}
+
+/// An `Angle` formed by the particles at indexes `i`, `j` and `k`
+///
+/// This structure ensure uniqueness of the `Angle` representation by enforcing
+/// `i < k`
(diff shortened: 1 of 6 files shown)


src/core/src/sys/connect.rs
```diff
@@ -0,0 +1,163 @@
+// Lumol, an extensible molecular simulation engine
+// Copyright (C) 2015-2016 G. Fraux — BSD license
+
+//! Data about bonds and angles in the system.
+use std::cmp::{min, max};
+
+/// A `Bond` between the atoms at indexes `i` and `j`
```
g-bauer: I'd go with `particles` instead of `atoms` in all descriptions. Makes it easier to understand since we also have the `Particle` struct.

src/core/src/sys/molecules.rs
```diff
@@ -2,141 +2,15 @@
 // Copyright (C) 2015-2016 G. Fraux — BSD license
 
 //! Data about bonds and angles in the system.
```
g-bauer: Data about molecules in the system

src/core/src/sys/molecules.rs
```diff
@@ -154,10 +28,8 @@ pub struct Molecule {
     /// the connectivity between the particles `i + self.first` and
     /// `j + self.first`
     connections: Array2<Connectivity>,
-    /// Index of the first atom in this molecule.
-    first: usize,
-    /// Index of the last (included) atom in this molecule.
-    last: usize,
+    /// Range of atomic indexes in this molecule.
+    atoms: Range<usize>,
```
g-bauer: Maybe rename this to something like `particles_range`? I think `particles` would also work but could lead to confusion.

src/core/src/sys/molecules.rs
```diff
@@ -176,39 +48,39 @@ impl Molecule {
...
     pub fn first(&self) -> usize {
-        self.first
+        self.atoms.start
     }
 
-    /// Get the last atom of this molecule
-    pub fn last(&self) -> usize {
-        self.last
+    /// Get the index of the first atom after this molecule
+    pub fn next_atom(&self) -> usize {
```
g-bauer: I think the name is misleading. This reads somewhat like an iterator, returning the next atom of  `self`.

Luthaf: How about `end` then, like the std `Range`? The point is that this value is the first index not in the molecule.

Luthaf:
Rebased and corrected.

The codecov status is weird, it looks like it is now considering closing braces as (sometimes not coverged) lines of there own, while it was not before. [before](https://codecov.io/gh/lumol-org/lumol/src/1645f56aa944ef4afbe212354fa0a45e31532ee3/src/core/src/out.rs#L125) - [after](https://codecov.io/gh/lumol-org/lumol/src/92808b4bb37c7cdb552d25ae2c3f4be88d46bda6/src/core/src/out.rs#L125). What should we do? I can try to add tests to bring coverage up, or we can relax the target value and set it to 84%.

g-bauer:
Well, I think - for now - we could just lower the coverage target. I think it would be good to specifically
 look at adding (good) tests before we go public. I'd like to merge, if you're ok with it. 👍

Luthaf:
> I'd like to merge, if you're ok with it. 👍

Just let me change coverage target before, and we'll be good.

PR #70, opened 2016-12-14
Add setup function for MCMove

g-bauer:
This PR adds a `setup` function for `MCMove`.

The interface is identical to the `setup` function for the `Integrator` trait:

```rust
fn setup(&mut self, system: &System) { ... }
```

This can be used to get information from the `System` before the simulation starts. For example, we can get the cutoff radius from the interaction potentials, which is useful to decide how much a cell can be resized or to limit the amplitude of translations.

For the currently implemented moves, `rotate` and `translate`, this function does nothing. For `translate` we might want to add a limitation for the translation range `dr` that can be set here.

### DIFF
--- a/src/core/src/sim/mc/monte_carlo.rs
+++ b/src/core/src/sim/mc/monte_carlo.rs
@@ -145,6 +145,9 @@ impl Propagator for MonteCarlo {
     fn setup(&mut self, system: &System) {
         self.normalize_frequencies();
         self.cache.init(system);
+        for mc_move in self.moves.iter_mut() {
+            mc_move.0.setup(system)
+        } 
     }
 
     fn propagate(&mut self, system: &mut System) {
@@ -194,6 +197,17 @@ impl Propagator for MonteCarlo {
             mcmove.1.nattempted = 0;
         }
     }
+
+    /// Print some informations about moves to screen
+    fn finish(&mut self, _: &System) {
+        println!("Monte Carlo simulation summary");
+        for mc_move in self.moves.iter() {
+            println!("Statistics for move: {}", mc_move.0.describe());
+            println!("  Calls     : {}", mc_move.1.ncalled);
+            println!("  Acceptance: {} %", mc_move.1.naccepted as f64 /
+             mc_move.1.nattempted as f64 * 100.0);
+        }
+    }
 }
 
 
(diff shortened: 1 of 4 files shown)


src/core/src/sim/mc/monte_carlo.rs
```diff
@@ -145,6 +145,9 @@ impl Propagator for MonteCarlo {
     fn setup(&mut self, system: &System) {
         self.normalize_frequencies();
         self.cache.init(system);
+        for mc_move in self.moves.iter_mut() {
```
Luthaf: `for mc_move in &mut self.moves` is more idiomatic.

src/core/src/sim/mc/monte_carlo.rs
```diff
@@ -194,6 +197,17 @@ impl Propagator for MonteCarlo {
             mcmove.1.nattempted = 0;
         }
     }
+
+    /// Print some informations about moves to screen
+    fn finish(&mut self, _: &System) {
+        println!("Monte Carlo simulation summary");
```
Luthaf: You should use `info!` here, as the standard output may not exist, or the user may want to redirect everything to a file.

src/core/src/sim/mc/monte_carlo.rs
```diff
@@ -315,6 +330,7 @@ mod tests {
         let mut mc = MonteCarlo::new(100.0);
         mc.add_move_with_acceptance(Box::new(DummyMove), 1.0, 0.5);
         mc.moves[0].1.set_acceptance(Some(1.1));
+        mc.moves[0].1.set_acceptance(Some(0.0));
```
Luthaf: You need one test by `panic!`, the two last line will never execute here.

src/core/src/sim/mc/moves/rotate.rs
```diff
@@ -114,7 +116,9 @@ impl MCMove for Rotate {
 
     fn update_amplitude(&mut self, scaling_factor: Option<f64>) {
         if let Some(s) = scaling_factor {
-            self.theta *= s;
+            if (s * self.theta).to_degrees() <= 180.0 {
+                self.theta *= s;
+            }
```
Luthaf: Maybe we can send a warning (using `warn_once!`) if the move amplitude would have gotten bigger than 180°.

src/core/src/sim/mc/moves/translate.rs
```diff
@@ -64,6 +64,12 @@ impl MCMove for Translate {
         "molecular translation"
     }
 
+    fn setup(&mut self, _: &System) { 
+        // Maybe introduce a limitation for dr, dr_max, such that it can not
```
Luthaf: Can you add a `TODO` here ?

Luthaf:
The codecov/patch value is OK here, we do not have unit tests for MC moves.

Thank you for the two PR !

PR #71, opened 2016-12-14
Add function to get the largest cutoff radius

g-bauer:
This PR adds functionality to access the largest cutoff from all `pairs`. This is useful to get the limit of the smallest simulation cell.

Currently, only pair interactions are considered, no electrostatics.

### DIFF
--- a/src/core/src/energy/pairs.rs
+++ b/src/core/src/energy/pairs.rs
@@ -147,6 +147,11 @@ impl PairInteraction {
     pub fn set_restriction(&mut self, restriction: PairRestriction) {
         self.restriction = restriction;
     }
+
+    /// Return the cutoff radius
+    pub fn get_cutoff(&self) -> f64 {
+        self.cutoff
+    }
 }
 
 impl PairInteraction {
@@ -304,6 +309,8 @@ mod tests {
 
         assert_eq!(pairs.force(4.1), 0.0);
         assert_eq!(pairs.energy(4.1), 0.0);
+
+        assert_eq!(pairs.get_cutoff(), 4.0);
     }
 
     #[test]
--- a/src/core/src/energy/pairs.rs
+++ b/src/core/src/energy/pairs.rs
@@ -149,6 +149,18 @@ impl PairInteraction {
     }
 
     /// Return the cutoff radius
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use lumol::energy::PairInteraction;
+    /// use lumol::energy::LennardJones;
+    ///
+    /// let ar = LennardJones{sigma: 3.405, epsilon: 1.0};
+    /// let interaction = PairInteraction::new(Box::new(ar), 9.1935);
+    /// let rc = interaction.get_cutoff();
+    /// assert_eq!(rc, 2.7f64 * 3.405);
+    /// ```
     pub fn get_cutoff(&self) -> f64 {
         self.cutoff
     }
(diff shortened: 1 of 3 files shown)


src/core/src/energy/pairs.rs
```diff
@@ -147,6 +147,11 @@ impl PairInteraction {
     pub fn set_restriction(&mut self, restriction: PairRestriction) {
         self.restriction = restriction;
     }
+
+    /// Return the cutoff radius
+    pub fn get_cutoff(&self) -> f64 {
```
Luthaf: As this is a public function, can you add an example here? See the other functions or Rust standard library for example of `examples` 😄 The example should use this function and have at least one `assert_eq!` for checking the function result.

src/core/src/sys/interactions.rs
```diff
@@ -185,6 +185,20 @@ impl Interactions {
         }
     }
 
+    /// Get all pair interactions
+    ///
+    /// Using this function, you lose information
+    /// about which interaction belongs to which pair of particles.
+    pub fn all_pairs(&self) -> Vec<&PairInteraction> {
+        let res: Vec<&PairInteraction> = self.pairs
+            .values() // value is a Vec<&PairInteraction>
```
Luthaf: `Vec<Vec<&PairInteractions>>`

src/core/src/sys/interactions.rs
```diff
@@ -185,6 +185,20 @@ impl Interactions {
...
+            // flatten the vector -> return an iterator over
+            // Vec<&PairInteraction>
+            .flat_map(|i| i)
+            .collect();
+        res
```
Luthaf: I believe this could be written

```rust
pub fn all_pairs(&self) -> Vec<&PairInteraction> {
    self.pairs.values() // value is a Vec<Vec<&PairInteraction>>
                        // flatten the vector -> return an iterator over
                        // Vec<&PairInteraction>
              .flat_map(|i| i)
              .collect()
}
```

src/core/src/sys/systems.rs
```diff
@@ -465,6 +464,19 @@ impl System {
         let mkind = self.particles[m].kind;
         self.interactions.dihedrals(ikind, jkind, kkind, mkind)
     }
+
+    /// Return the largest cutoff radius from pair interactions
+    /// excluding electrostatic interactions.
+    pub fn largest_cutoff(&self) -> f64 {
```
Luthaf: I am not comfortable with having this as a `System` function. It feels like it belongs to `Interactions` instead, and should be accessed by `system.interactions().largest_cutoff()`

`System` already have too much functions =/

Aside, how often do you plan to use this ? If it is only for `Resize`, it could just be a piece of code inside resize, that we can refactor in a public function later if needed.

g-bauer:
I updated the comments and added an example for `get_cutoff`.

Also, I agree that a `fn largest_cutoff` is not needed since it will not be called too often and therefore it should be computed explicitly for each respective case. Sadly, the name of the PR does not capture the actual changes anymore.

src/core/src/sys/interactions.rs
g-bauer: I am not sure here. I thought that the value of the tree is a `Vec<PairInteraction>` looking at the definition in the struct. `flat_map` produces an iterator over the elements, i.e. `PairInteraction`s. This is done for every entry in the tree and finally everything is collected into a single vector. Did I get this wrong?

Luthaf:
> Sadly, the name of the PR does not capture the actual changes anymore. 

This is not too important, and you can change the PR name !

Issue #64, opened 2016-12-03
Drift in energy caused by energy cache


g-bauer:
I observed a drift in energy for long simulations of LJ particles in the NVT ensemble.

When a move is accepted, I compared the energy delta (energy_new - energy_old) from cache via 
`du_cache = cache.move_particles_cost(...)` versus the difference computed via `system.total_energy()`. Looking at the difference, I found that often the difference is negligible (1e-16) but there are cases when the difference is about 1e-7 to 1e-5.

I think the problem is either `move_particles_cost` or the update function of the cache. As far as I can tell, the initialization of the cache works correctly (tail correction is missing at the moment).

You can find the results from a sample simulation [here](https://github.com/g-bauer/jupyter_notebooks/blob/master/cache_energy_drift.ipynb).

Luthaf:
Fixed by #72

PR #72, opened 2016-12-16
Debug cache: move_particle_cost

g-bauer:
This PR addresses two topics:

### 1) Fix the bug from #64 

The bug was in the update closure [here](https://github.com/lumol-org/lumol/blob/12ebbf6379da4178eb4ca62a866a04f762979f18/src/core/src/sys/cache.rs#L245).

Imagine particle `i` moves away from particle `j` so that the old distance is `r_old < rc` and the new distance is `r_new > rc`. It's now at a distance larger than the cutoff which will lead to `energy_old != 0.0 -> energy_new = 0.0`. In the current state, this true energy will be skipped due to the if-clause. 

See [this](https://github.com/g-bauer/jupyter_notebooks/blob/master/cache_energy_debugged.ipynb) notebook for a simulation with the changes.

### 2) Add tail corrections to the cache.

### DIFF
--- a/src/core/src/sys/cache.rs
+++ b/src/core/src/sys/cache.rs
@@ -162,21 +162,6 @@ impl EnergyCache {
             }
         }
 
-        // Interactions withing the sub-system being moved
-        for (i, &part_i) in idxes.iter().enumerate() {
-            for (j, &part_j) in idxes.iter().enumerate().skip(i + 1) {
-                let r = system.cell().distance(&newpos[i], &newpos[j]);
-                let energy = evaluator.pair(r, part_i, part_j);
-
-                pairs_delta += energy;
-                new_pairs[(part_i, part_j)] += energy;
-                new_pairs[(part_j, part_i)] += energy;
-
-                pairs_delta -= self.pairs_cache[(part_i, part_j)];
-            }
-        }
-
-
         // Recompute everything here. One day we could store some terms and do
         // the same thing than for pair interactions.
         let mut bonds = 0.0;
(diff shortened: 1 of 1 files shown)


src/core/src/sys/cache.rs
```diff
@@ -162,21 +162,6 @@ impl EnergyCache {
...
-                let energy = evaluator.pair(r, part_i, part_j);
-
-                pairs_delta += energy;
-                new_pairs[(part_i, part_j)] += energy;
-                new_pairs[(part_j, part_i)] += energy;
-
-                pairs_delta -= self.pairs_cache[(part_i, part_j)];
-            }
-        }
-
```
g-bauer: This may be reintroduced in the next step of this PR. Currently, I can think of no scenario where the intra and intermolecular structure of multiple (but not all) particles/molecules may change. If all change, like in volume changes, we should write an own caching function.

Luthaf: I think I wrote this code to handle the case of flexible molecules. As it is tested (removing it broke the tests), we can just keep it around.

g-bauer: Yes, I just put it in back.

Luthaf:
> Imagine particle i moves away from particle j so that the old distance is r_old < rc and the new distance is r_new > rc. It's now at a distance larger than the cutoff which will lead to energy_old != 0.0 -> energy_new = 0.0. In the current state, this true energy will be skipped due to the if-clause.

And this is an impressive catch !

g-bauer:
... and it was a bit tedious to find 😄

PR #73, opened 2016-12-17
Expose Minimization in input

Luthaf:
This PR extracts a Minimizer trait, similar to the propagator trait, and then expose the minimization capacities in the input files.

The minimization assumes that all Minimizer can be used with an energy/force criterion for minimization convergence, and uses the [following algorithm](https://github.com/lumol-org/lumol/commit/47580000fa8b9d2a277eee4d1b2fedd608fda555#diff-80d98947cebb0c4b8f6df85d66dba954R82) to check for convergence. The main function here is [`Minimizer::minimize`](https://github.com/lumol-org/lumol/commit/47580000fa8b9d2a277eee4d1b2fedd608fda555#diff-80d98947cebb0c4b8f6df85d66dba954R29), that should generate a new configuration of possibly lower energy, and return the corresponding energy/force pair to allow `Minimization` to check for convergence.

Does this design look good to you?

### DIFF
--- a/doc/src/SUMMARY.md
+++ b/doc/src/SUMMARY.md
@@ -23,6 +23,7 @@
     - [Simulations](input/simulations.md)
     - [Molecular dynamics](input/md.md)
     - [Monte-Carlo](input/mc.md)
+    - [Minimization](input/min.md)
 
 - [Advanced tutorials]()
     - [Adding potentials]()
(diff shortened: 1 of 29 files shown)


Luthaf:
Done !

Luthaf:
Also, ping @odarbelaeze, since you asked about this on gitter the other day.

g-bauer:
Awesome, I'd merge if there are no additional comments.

PR #74, opened 2017-01-01
Specify the linker for windows cross compilation

odarbelaeze:
Hi there, it's nice to provide executables for multiple platforms and I just realised rust makes this dead easy. However, it seems like when you execute a cross compilation command:
```
cargo build --target x86_64-pc-windows-gnu
```
it tries to use the normal `gcc` as the linker, so I added this bit of config to have the cross compilation work for windows.

Let me know if you want me to document de cross compilation process a bit more in depht.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,5 @@
+[target.i686-pc-windows-gnu]
+linker = "i686-w64-mingw32-gcc"
+
+[target.x86_64-pc-windows-gnu]
+linker = "x86_64-w64-mingw32-gcc"


Luthaf:
I don't know if we want to include such configuration in the main repo. As I understand the `.cargo/config` files hierarchy, you could accomplish the same effect by moving this file to `~/.cargo/config`.

My issue here is that we force the linker name to be `i686-w64-mingw32-gcc` on all systems, and I am not sure it will always have this name (it could be `i386-w32-mingw32-gcc` for example).

Nevertheless, we could add documentation somewhere with this exact file as example, as it should cover the most common needs (I believe you are cross-compiling from Linux, is that right?).

> Let me know if you want me to document de cross compilation process a bit more in depht.

As a side question, you did manage to cross-compile and run the whole project? That would be really nice and an additional proof of the greatness of the `cmake/gcc` crates, as one of the dependency of lumol is written in C++.

odarbelaeze:
@Luthaf Yeah, you're right, some docs about how to cross compile the project will suffice, and yes, the cross compilation created a `lumol.exe` executable, I got a windows dual bot so I can check if the executable actually works.

Luthaf:
Ok, I am closing this PR, and will welcome something improving the documentation!

Issue #76, opened 2017-01-06
`Trajectory::open` shows different behavior depending on the order of atoms


g-bauer:
I try to read three configurations with a single water molecule from `xyz` files. Depending on the order, it either works or throws an error. I simply changed the order by copying the whole line (coordinates are consistent) but it should not matter. 

To reproduce, copy `.rs` file from [this gist](https://gist.github.com/g-bauer/883b0b34460cbcaee0f3199c0dd55a4d) into the example folder and `.xyz` into example/data.

Error:
```bash
$ RUST_BACKTRACE=1 cargo run --example=error_molid --release
thread 'main' panicked at 'assertion failed: new_molid < old_molid', lumol/src/core/src/sys/systems.rs:342
stack backtrace:
   1:        0x1095141e8 - std::sys::backtrace::tracing::imp::write::h6f1d53a70916b90d
   2:        0x10951637f - std::panicking::default_hook::{{closure}}::h137e876f7d3b5850
   3:        0x109515475 - std::panicking::default_hook::h0ac3811ec7cee78c
   4:        0x109515a26 - std::panicking::rust_panic_with_hook::hc303199e04562edf
   5:        0x1094c04f4 - std::panicking::begin_panic::h341b039f84d0b176
   6:        0x1094c9041 - lumol::sys::systems::System::add_bond::ha262c429a07c9463
   7:        0x1094ca985 - <chemfiles::frame::Frame as lumol::sys::chfl::ToLumol>::to_lumol::h86945c7802c5c377
   8:        0x1094cac6a - lumol::sys::chfl::Trajectory::read_guess_bonds::hb213dbb19c866ab4
   9:        0x1094bedd3 - error_molid::main::h83ddaaca3f875982
  10:        0x10951693a - __rust_maybe_catch_panic
  11:        0x109514fb6 - std::rt::lang_start::h538f8960e7644c80
```

Luthaf:
I believe this bug comes from the already complex code that deals with merging molecules when adding new bonds. I'll give it a look.

PR #77, opened 2017-01-07
Re-order molid before merging molecules

Luthaf:
A simple fix for #76: the assertion was documented but not checked by all the callers, I changed merge_molecule to reorder molid as it need.

### DIFF
--- a/src/core/src/sys/systems.rs
+++ b/src/core/src/sys/systems.rs
@@ -319,16 +319,17 @@ impl System {
         self.particles.iter_mut()
     }
 
-    /// Merge the molecules at indexes `new_molid` and `old_molid` into one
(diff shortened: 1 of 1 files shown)


g-bauer:
Looks good to me. Can you check the example (i cannot add a comment at that line) in the function documentation?

```rust
     /// H-H  H-H  H-H  H-H
     /// 0 1  2 3  4 5  6 7
     /// ```
     /// and call `merge_molecules(0, 2)` the result will be
```
Should this be `merge_molecules(0, 3)`?

Luthaf:
> Should this be merge_molecules(0, 3)?

Yes, nice catch.

g-bauer:
Merged. 👍

PR #80, opened 2017-01-16
[DO NOT MERGE] Generate struct of array version of Vec<Particle>

Luthaf:
This is an initial and very rough proof-of-concept for using procedural macro to automatically generate all the helper functions and structs needed to do an Array of structs => struct of arrays automatically. For more background on this transformation and why it can be useful, see https://github.com/lumol-org/lumol/issues/18#issuecomment-272833695, and https://en.wikipedia.org/wiki/AOS_and_SOA.

This implementation uses the not-yet-stable procedural macros feature, so you need the beta version of the compiler to compile/test it:

```
rustup toolchain add beta
cd /path/to/lumol
cd src/macros
cargo +beta test
```

---

The code currently generate a single `ParticleVec` struct from a `Particle` definition; I would like to generate equivalents to `&Particle`; `&mut Particle`; `&[Particle]` and `&mut [Particle]` too. It only contains two methods: `push` and `len`, we need to look at all the `Vec` and `[T]` methods to know which one we need to replicate here.

We can not just use `Deref<Target=Vec<Particle>>`, because this is not how the data is laid out in memory.

Here would be an example of which structs would be generated in the [example](https://github.com/lumol-org/lumol/compare/soa-macros?expand=1#diff-e291cfad782d1f093307dfca338f363fR4).
```rust
// Same as &'a Particle, with data inside the ParticleVec
struct ParticleRef<'a> {
    pub name: &'a String,
    pub mass: &'a f64
}

// Same as &'a mut Particle, with data inside the ParticleVec
struct ParticleMut<'a> {
    pub name: &'a mut String,
    pub mass: &'a mut f64
}

// Same as &'a [Particle], with data inside the ParticleVec
struct ParticleSlice<'a> {
    pub name: &'a [String],
    pub mass: &'a [f64]
}

// Same as &'a mut [Particle], with data inside the ParticleVec
struct ParticleSliceMut<'a> {
    pub name: &'a  mut [String],
    pub mass: &'a mut [f64]
}
```

---

Before being merged, this will need us to gather more data on speed gain or loss, and to run more benchmarks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,8 @@ license = "BSD"
 [workspace]
 members = [
 	"src/core",
-    "src/input"
+    "src/input",
+    "src/macros",
 ]
 
 [[bin]]
--- a/src/macros/Cargo.toml
+++ b/src/macros/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "lumol-macros"
+version = "0.1.0"
+authors = ["Luthaf <luthaf@luthaf.fr>"]
+documentation = "https://lumol-org.github.io/lumol"
+repository = "https://github.com/lumol-org/lumol"
+readme = "../../README.md"
+license = "BSD"
+
+[dependencies]
+syn = "0.10"
+quote = "0.3"
+
+[lib]
+proc-macro = true
--- a/src/macros/src/lib.rs
+++ b/src/macros/src/lib.rs
@@ -1,0 +1,81 @@
+// Lumol, an extensible molecular simulation engine
+// Copyright (C) 2015-2016 G. Fraux — BSD license
+extern crate proc_macro;
+extern crate syn;
+#[macro_use]
+extern crate quote;
+
+use syn::{Body, VariantData, MacroInput};
+use proc_macro::TokenStream;
+
+#[proc_macro_derive(StructOfArray)]
+pub fn soa_derive(input: TokenStream) -> TokenStream {
+    let source = input.to_string();
+    let ast = syn::parse_macro_input(&source).unwrap();
+    let generated = derive_vec(&ast);
+    generated.parse().unwrap()
+}
+
+fn derive_vec(input: &MacroInput) -> quote::Tokens {
+    let ident = &input.ident;
+    let soa_ident = quote::Ident::from(format!("{}Vec", input.ident));
+    let fields = match input.body {
+        Body::Struct(ref data) => {
+            match data {
+                &VariantData::Struct(ref fields) => fields.clone(),
+                _ => panic!("#[derive(SoA)] only supports structs."),
+            }
+        }
+        _ => panic!("#[derive(SoA)] only supports structs."),
+    };
+
+    let vec_fields = fields.iter()
+        .map(|f| {
+            let field_ident = f.ident.clone().unwrap();
+            let field_ty = &f.ty;
+            quote!{
+                pub #field_ident: Vec<#field_ty>
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let field_idents = fields.iter().map(|f| f.ident.clone().unwrap()).collect::<Vec<_>>();
+    let field_idents_1 = &field_idents;
+    let field_idents_2 = &field_idents;
+
+    let first_ident = &field_idents[0];
+
+    quote!{
+        #[derive(Debug)]
+        struct #soa_ident {
+            #(
+                #vec_fields,
+            )*
+        }
+
+        impl #soa_ident {
+            pub fn new() -> Self {
+                #soa_ident {
+                    #(
+                        #field_idents_1 : Vec::new(),
+                    )*
+                }
+            }
+
+            pub fn push(&mut self, value: #ident) {
+                let #ident{#(#field_idents_1),*} = value;
+                #(
+                    self.#field_idents_1.push(#field_idents_2);
+                )*
+            }
+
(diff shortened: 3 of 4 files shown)


g-bauer:
That's neat, although I (currently) find the macros to create the code rather complex. But I did not work with macros too much yet.

In this form, we have to rewrite access patterns to the data, i.e `particle[i].position` will be `particle.position[i]`. I cannot think of a way to make an unified interface to access the data (but I still only scratch on the surface of what rust is capable of) - maybe there is a way to do this?

If this would be possible, one could write functions that take either SoA or AoS type structs as arguments, the body of the function performs work on both structs in the same way, since access is unified. The advantage would be that switching the memory layout (SoA <-> AoS) would not involve any changes to the main code, just the declaration. But I think that is not possible in rust - and for the scope of this project, we are not interested in switching the format.

Luthaf:
> That's neat, although I (currently) find the macros to create the code rather complex. But I did not work with macros too much yet.

The `quote!` syntax looks a bit like the `macro_rules!` one, so you can have a look at it.

> In this form, we have to rewrite access patterns to the data, i.e particle[i].position will be particle.position[i]

We effectively can not have `particles[i].position`, because `Index::index` is required to give a reference, and we can not build one. But we can have ` particles.get(i).position`, with `ParticleVec::get(&'a self, i: usize) -> ParticleRef<'a>`. I do not know if the optimizer is good enough to remove the construction/destruction of `ParticleRef` here, but I hope so.

> If this would be possible, one could write functions that take either SoA or AoS type structs as arguments, the body of the function performs work on both structs in the same way, since access is unified.

I have no idea about how you could write this, I'll try to think about it. I am not sure whether it would be useful in our case anyway, except for benchmarks.

---

Another possibility I want to explore here is the ability to iterate over subsets of the data with a nice API. I'd like to be able to write:

```rust
// Use a subset of properties
for (position, velocity, mass) in particles.physical() {

}

// another predefined subset
for (name, kind) in particles.kind_data() {

}

// Dynamically defined subset of properties
for (position, name) in particles.properties([Properties::Position, Properties::Name]) {

}
```

This could improve readability, but will need to be benchmarked to be sure it is not worst that the others options.

g-bauer:
```rust
// Use a subset of properties
for (position, velocity, mass) in particles.physical() {

}
```

For that, we may use [`multizip` from `itertools`](https://docs.rs/itertools/0.5.8/itertools/fn.multizip.html).

```rust
extern crate itertools;
use itertools::multizip;

// p is a ParticleVec with some fields including `mass`, `position` and `velocity`.
for (mass, position, velocity) in multizip((&p.mass, &p.position, &p.velocity))  {
    println!("Mass: {:?}", mass);
    println!("Position: {:?}", position);
    println!("Velocity: {:?}", velocity);
    println!("");
}
```

I think we could write our own functions (e.g. `physical`) that return the iterator created via `multizip`. This may be slow since (i guess) it will checks bounds for every single iterator (here: three).

Luthaf:
>  This may be slow since (i guess) it will checks bounds for every single iterator (here: three).

The author of itertools wrote somewhere (i can't remember where) that you could prevent bound checking in some cases by explicitly creating slices with the right length:

```rust
extern crate itertools;
use itertools::multizip;

// particles is a ParticleVec with some fields including `mass`, `position` and `velocity`.
let len = particles.len();
let masses = &particles.mass[..len];
let positions = &particles.position[..len];
let velocities = &particles.velocity[..len];

for (mass, position, velocity) in multizip((masses, positions, velocities))  {
    println!("Mass: {:?}", mass);
    println!("Position: {:?}", position);
    println!("Velocity: {:?}", velocity);
    println!("");
}
```

I do not know if this holds for `multizip`, and if this helps when hiding code behind a function (maybe with `#[inline]`?)

Luthaf:
Throwing more ideas in, we could have two arrays instead of one here: a `cold` array containing the atom name and any other properties not accessed that often; and a `hot` array containing the position, velocity, mass, ...

```rust
ParticleCold {
     name: String,
}

ParticleHot {
    position: Vector3D,
    velocity Vector3D,
    mass: f64,
}

ParticleVec {
    hot: Vec<ParticleHot>,
    cold: Vec<ParticleCold>
}
```

This way, the things that are accessed together are grouped together in memory, while still preventing cache bloat with useless data.

g-bauer:
I'd love to have a way to pull the namespace out of a struct - here to make fields of both `hot` and `cold` directly accessible. If `ParticleVec` could access both structs' fields names (so that we could write `ParticleVec.field` instead of `ParticleVec.hot.field`) we could easily shuffle the fields around to benchmark what to put where. 

It is very hard to tell how to do the separation. For MC, you would use only positions (and maybe the particle kind). For MD, most of the time is spent in force computation, which also only needs positions and kinds to determine the potential functions. Updating positions on the other hand needs velocities and masses. 

```rust
pub struct Particle {
    // identifier, not used during simulation (?)
    name: String,
    // energy, force -> potential lookup
    pub kind: ParticleKind,
    // position updates in MD, com     
    pub mass: f64,
    // Ewald/Wolf energy, force                  
    pub charge: f64,
    // energy, force, com               
    pub position: Vector3D,  
     // position updates in MD  
    pub velocity: Vector3D,  
}
```

Luthaf:
I extracted this code in [soa-derive](https://github.com/lumol-org/soa-derive), as it is not at all dependant to lumol internals. I'll finish coding there, and then open a new PR with benchmarks results!

Issue #63, opened 2016-12-02
Replace the warn_once! macros by the log-once crate


Luthaf:
I pulled the code from [here](https://github.com/lumol-org/lumol/blob/47665eed673692e1bed2b741c01ea21e0855c680/src/core/src/logging.rs#L39-L57) to its own crate: [log-once](https://crates.io/crates/log-once). We should remove the in-tree macro and replace it with the crate.

PR #81, opened 2017-01-17
Use warn_once! macro from external lib

antoinewdg:
Issue #63.


Luthaf:
Thank for the PR!

Can you check if this also [fix this](https://github.com/lumol-org/lumol/blob/1645f56aa944ef4afbe212354fa0a45e31532ee3/src/core/src/sim/mc/moves/rotate.rs#L123)? I believe this was a Clippy lint warning that got triggered by the previous code, but as the implementation changed, it may not be relevant any more.

antoinewdg:
Not sure I'm getting what was the issue here. Was there a compiler warning caused by the macro `wanr_once!` that should not happen anymore with the macro from `log-once` ?

Luthaf:
> Not sure I'm getting what was the issue here. 

[Clippy](https://github.com/Manishearth/rust-clippy) is an additional collection of lints (static analysis) that one can opt-in to use with rust project. One of these lint was sending a warning on this code.

You can install clippy using:
```
rustup toolchain add nightly
cargo +nightly install clippy
```

And run it on the code using:
```
cd path/to/lumol
cargo +nightly clippy
```

> Was there a compiler warning caused by the macro warn_once! that should not happen anymore with the macro from log-once ?

It could have been, but I actually still get it:

```
$ cargo +nightly clippy
    Updating registry `https://github.com/rust-lang/crates.io-index`
 Downloading log-once v0.1.0
   Compiling lazy_static v0.2.2
   Compiling log v0.3.6
   Compiling matrixmultiply v0.1.13
   Compiling bitflags v0.7.0
   Compiling rawpointer v0.1.0
   Compiling gcc v0.3.41
   Compiling either v1.0.2
   Compiling libc v0.2.18
   Compiling num-traits v0.1.36
   Compiling ndarray v0.6.10
   Compiling itertools v0.5.8
   Compiling log-once v0.1.0
   Compiling rand v0.3.15
   Compiling special v0.7.3
   Compiling num-complex v0.1.35
   Compiling cmake v0.1.20
   Compiling chemfiles-sys v0.6.0
   Compiling chemfiles v0.6.1
   Compiling lumol-core v0.1.0 (file:///.../lumol/src/core)
warning: useless use of `format!`, #[warn(useless_format)] on by default
   --> src/sim/mc/moves/rotate.rs:123:17
    |
123 |                   warn_once!(
    |  _________________^ starting here...
124 | |                     "Tried to increase the maximum amplitude for rotations to more than 180°."
125 | |                 );
    | |__________________^ ...ending here
    |
    = help: for further information visit https://github.com/Manishearth/rust-clippy/wiki#useless_format
    = note: this error originates in a macro outside of the current crate

    Finished dev [unoptimized + debuginfo] target(s) in 18.12 secs
```

antoinewdg:
Thanks ! I actually get a ton of warnings doing that, how come you don't have them ?

Luthaf:
Because I am running cargo clippy from `src/core`. There are a lot of clippy lint, as we use a specific configuration for it [here](https://github.com/lumol-org/lumol/blob/dbe819f52142470b94bce84472ff768fb8da5374/src/core/src/lib.rs#L19-L34). But this configuration is limited to the `lumol_crate`, as there is no way to setup the lint level in a global configuration file for clippy yet. So we get quite a lot of warnings in the others files, mainly the tests and examples.

antoinewdg:
OK my bad, I have the same output then, I reverted the commit.
Is there a reason why the clippy lint is not run on Travis ?

Luthaf:
>  Is there a reason why the clippy lint is not run on Travis?

No special reason, except that we do not want to impose to much pressure on potential contributors about external tools that are needed locally. But this can change, given we are sure that we want that. Currently, I am running clippy on the code every months or two, which was OK for now.

Issue #31, opened 2016-07-05
Ability to run control algorithm not at all steps


Luthaf:
Control algorithms like `RemoveRotation` and `RemoveTranslation` do not need to run at each simulation step. they could have separated frequency field to run them only every so often.

antoinewdg:
Is there any docs about how `RemoveRotation` and `RemoveTranslation` should be described in input files ?

antoinewdg:
My bad, [found them](http://lumol-org.github.io/lumol/latest/book/input/md.html).

antoinewdg:
So how should this change be reflected in the TOML file ?
Should there be a `period` key for each control entry ? Should this key be optional ? Default to 1 ?

Or do we just consider that this can not be controlled by the user and set it to a reasonable default ?

g-bauer:
I'd prefer to expose this to the user, but optional. What do you think about this: 

```toml
[simulations.propagator]
type = "MolecularDynamics"
timestep = "1 fs"
thermostat = {type = "Berendsen", temperature = "400 K", timestep = 100}
controls = [
    {type = "RemoveRotation", timestep = 100}
]
```

This would be in line with the thermo- and barostat that also use `timestep` (as factor for the "real" timestep). Now that I am writing this, using two `timestep` could be confusing ...

Luthaf:
> This would be in line with the thermo- and barostat that also use timestep (as factor for the "real" timestep). Now that I am writing this, using two timestep could be confusing.
>
> How about after_steps = 100, steps = 100?

The Monte-Carlo moves uses [`frequency`](https://github.com/lumol-org/lumol/blob/dbe819f52142470b94bce84472ff768fb8da5374/src/input/tests/simulation/good/mc.toml#L15), which I prefer.

antoinewdg:
@g-bauer well I kind of feel that using `timestep` is already super confusing for `Berendsen` and such, since this is not a time step as for the `Propagator`, this is a scaling factor to apply to the `timestep` given to the `Propagator`. 
Maybe this is on account of me not having any background in molecular dynamics, but this felt really confusing. I fist thought it just overrode the `Propagator` `timestep`, is there a particular reason for calling both of them `timestep`? Why not just `tau` as in the code ?

@Luthaf why not `frequency`, this would have to be between 0 and 1 to make sense though, so it would not really be comparable to the `frequency` for MC move, which is an unormalized frequency.

Luthaf:
[...]

PR #84, opened 2017-01-18
Ability to run control algorithm not at all steps

antoinewdg:
Issue #31.
This PR:
* adds an `Alternator` helper struct to wrap some algorithms and call them only every N time 
* implements the wrapping for `RemoveRotation` and `RemoveTranslation` controls


antoinewdg:
This is a first draft, right now only implemented for `RemoveRotation`. 
* I used an `Alternator` struct to factor some (admiteddlyvery simple) code out of the `RemoveRotation`, I was pretty uninspired for naming and am open to any suggestion in that regard
* I  know where to put the `Alternator` so I put it in `utils`
* I set the default `frequency` to 1 because this is backward compatible 

By the way we need some kind of `assert_approx_eq!` for `Vector3D`.

antoinewdg:
TODO: 
- [ ] implement for `RemoveTranslation`
- [ ] add some doc in the code
- [ ] maybe update the book on this part ?
- [ ] add unit test for `Alternator` and the new TOML input

antoinewdg:
This is starting to look like something. Are we set on the name `every` for the TOML input ?
Is it normal that the `from_toml` functions are not unit tested and only tested through the big TOML files in `inputs/test` ?

doc/src/input/md.md
```diff
@@ -169,16 +169,17 @@ thermostat = {type = "Rescale", temperature = "250 K", tolerance = "10 K"}
...
+by giving a control `type`. The `every` key specifies that the algorithm should
+only be run every `n` step of the simulation (optional, defaults to 1).
 
 ```toml
 [simulations.propagator]
 type = "MolecularDynamics"
 timestep = "1 fs"
 controls = [
-    {type = "RemoveRotation"}
+    {type = "RemoveRotation", every = 4}
```
g-bauer: What do you think about adding a comment here to make this very clear?
I really like these small comments in examples.

```toml
controls = [
# Remove rotation of the system every 4 fs
    {type = "RemoveRotation", every = 4}
]
```

g-bauer:
> Are we set on the name every for the TOML input ?

Yes from me.

Luthaf:
> I know where to put the Alternator so I put it in utils

There is also the [`OutputFrequency`](https://github.com/lumol-org/lumol/blob/112bf80eb76f34949e150550c6bb1d51949bb064/src/core/src/sim/simulations.rs#L11-L17) struct that does the same job, but in reverse: it wraps something existing, and run it only on specific steps. The two structs could be combined somehow. Maybe some approach using generics could work? I don't have a better idea for the name.

I prefer the wrapping approach, because the `Alternator` needs to be included in every struct that uses it, while the wrapping one can be used even for algorithms (controls or outputs) written outside of the core module.

What do you think here?

> Are we set on the name every for the TOML input ?

👍 

> Is it normal that the from_toml functions are not unit tested and only tested through the big TOML files in inputs/test ?

I could not think of a nice way to write unit tests here, so I went for comprehensive integration tests, checking that we actually get the right kind of error with the right error message. (see `src/input/tests/simulation/bad`). If you have an idea about how we can improve this, go for it!

> By the way we need some kind of assert_approx_eq! for Vector3D.

I wanted this too. Maybe we could define (or use from a crate) an `ApproxEq` trait like this:

```rust
trait ApproxEq<Rhs>: PartialEq<Rhs> {
    type Tolerance;
    fn approx_eq(&self, other: &Rhs, tolerance: Tolerance) -> bool;
    fn approx_eq_relative(&self, other: &Rhs, tolerance: Tolerance) -> bool;
}

impl ApproxEq<f64> for f64 {
    type Tolerance = f64;
    fn approx_eq(&self, other: &f64, tolerance: f64) -> bool {
        f64::abs(self - other) < tolerance
    }

    fn approx_eq_relative(&self, other: &f64, tolerance: f64) -> bool {
        f64::abs((self - other)/self) < tolerance
    }
}
```

doc/src/input/md.md
antoinewdg: Sure ! Maybe writing `every 4 timestep` would be clearer ?I feel that `every 4 fs` could lead one to believe they are expected to set a time in fs rather than a number of timesteps.

antoinewdg:
> I prefer the wrapping approach, because the Alternator needs to be included in every struct that uses it, while the wrapping one can be used even for algorithms (controls or outputs) written outside of the core module.

I agree. What do you think about something like that ?
``` rust
pub struct Alternator<T> {
    period: u64,
    control: T
}

impl<T> Control for Alternator<T> where T: Control {

    fn control(&mut self, system: &mut System) {
        if system.step() % self.period == 0 {
            self.control.control(system)
        }
    }
}

pub type RemoveTranslation = Alternator<BaseRemoveTranslation>;

```

We would just have to do the same kind of `impl` for different traits if we need to. About the naming, I'm still really bothered by anything using `frequency` to name what is actually a period. I don't want to sound annoying, but this seems a big deal to me.  

> I could not think of a nice way to write unit tests here, so I went for comprehensive integration tests, checking that we actually get the right kind of error with the right error message. (see src/input/tests/simulation/bad). If you have an idea about how we can improve this, go for it!

Right now this is probably a good approach, I just fear that when the configuration files get more complicated having every possible configuration will be difficult in integration tests, while it would be more doable with unit tests. I'll see what I can think about that would not be too painfull to write each time.

>  > By the way we need some kind of assert_approx_eq! for Vector3D.
>
> I wanted this too. Maybe we could define (or use from a crate) an ApproxEq trait like this:
...

Yep

doc/src/input/md.md
g-bauer: You are right, using `timestep` is better.

antoinewdg: Done :)

Luthaf:
> What do you think about something like that ?

I like it, except I would not use the type alias, but rather export both and allow the users to use create the `Alternator<RemoveRotation>` if they need; and use it ourself in `Simulation` or `MolecularDynamics`.

antoinewdg:
> I like it, except I would not use the type alias, but rather export both and allow the users to use create the Alternator<RemoveRotation> if they need; and use it ourself in Simulation or MolecularDynamics.

That sounds good. I'm almost there, I just still don't know where to put the `Alternator` struct. Right now it is in `sim/md/controls.rs`, but this doesn't seem appropriate since we may want to use it elsewhere. The `lumol::utils` module is private so it's not appropriate either. Maybe in `lumol::types` ?

Luthaf:
You can create a new module if you feels like there is a need for it. I would make it accessible as `sim::Alternator`, and put it somewhere close (either in the mod.rs file or in a sim/utils.rs module).

antoinewdg:
Well this is done, what do you guys think ?

src/core/src/sim/md/controls.rs
```diff
@@ -104,9 +105,27 @@ impl Control for BerendsenThermostat {
...
+    }
+}
+
 /// Remove global translation from the system
 pub struct RemoveTranslation;
 
+impl RemoveTranslation {
+    /// Create a new `RemoveTranslation"` control, that will
+    /// run every `every` step of the simulation.
+    pub fn new() -> RemoveTranslation {
```
Luthaf: Why do we need this? For the input, can't we just write `Alternator::new(every, RemoveRotation)`?

src/core/src/sim/mod.rs
```diff
@@ -15,3 +15,6 @@ pub use self::simulations::Simulation;
 pub use self::md::MolecularDynamics;
 pub use self::mc::MonteCarlo;
 pub use self::min::Minimization;
+
+mod utils;
+pub use self::utils::Alternator;
```
Luthaf: Please add a new line at the end of the file.

src/core/src/sim/utils.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /// Check if this is the appropriate time to run the algorithm.
+    pub fn can_run(&mut self) -> bool {
+        self.count += 1;
+        self.count % self.every == 0
+    }
+
+    /// Access the base algorithm.
+    pub fn base(&self) -> &T {
```
Luthaf: How about `as_ref`?

src/core/src/sim/utils.rs
```diff
@@ -0,0 +1,70 @@
...
+        &self.base
+    }
+
+    /// Access the base algorithm.
+    pub fn base_mut(&mut self) -> &mut T {
```
Luthaf: How about `as_mut`?

src/core/src/sim/utils.rs
Luthaf: Also, should we implement `Defer<Target=T>` and `DeferMut<Target=T>` too?

src/core/src/sim/md/controls.rs
antoinewdg: We probably can. However this is easier for backwards compatibility: if we ever need to add some fields to the struct, using only `RemoveTranslation` will break, while `RemoveTranslation::new()` could set some reasonable defaults.

src/core/src/sim/utils.rs
antoinewdg: Why not implementing `Defer` and `DeferMut`, but in this case no need for `as_ref` right ?

antoinewdg: Actually after thinking about it, I think that due to [coercions onmethod calls](https://doc.rust-lang.org/book/deref-coercions.html#deref-and-method-calls) implementing `Deref` could be a little confusing for no real benefit: `self.control()` and `(*self).control()` would be two different methods.

So I'm slightly in favor of `as_ref` and `as_ref_mut`.

Luthaf: > So I'm slightly in favor of as_ref and as_ref_mut.

Let's got for this.

src/core/src/sim/md/controls.rs
Luthaf: For the backward compatibly point of view, nothing prevent anyone from writing `let a = RemoveTranslation;`. We could have a struct with a private empty field (`struct RemoveTranslation(());`), but I am not sure this would be a win. Do we have any example of unit structs and backward compatibility best practises?

antoinewdg: I don't think we have. This is not really important, if the `new()` bothers you I can remove it.

src/core/src/sim/utils.rs
antoinewdg: Done

src/core/src/sim/mod.rs
antoinewdg: Done

src/core/src/sim/md/controls.rs
Luthaf: No that's OK.

We will need a plan for backward compatibility before reaching 1.0, we can discuss all this in a separated issue.

PR #88, opened 2017-01-22
Fix some examples

antoinewdg:
* the `xenon` example did not use the correct trajectory file
* are the paths in the `toml` files supposed to be relative to the file ? If so I fixed them, if not we have a problem because these examples do not run


Luthaf:
> are the paths in the toml files supposed to be relative to the file ?

Yes, they should.

PR #90, opened 2017-01-23
Add .editorconfig file

antoinewdg:
Ripped from the [crates.io repo](https://github.com/rust-lang/crates.io/blob/master/.editorconfig), this should add more consistency to the formatting (most editors support it), and in particular eliminate the "new line at the end of file" issue.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
--- a/Contributing.md
+++ b/Contributing.md
@@ -61,6 +61,11 @@ cd lumol
 cargo build
 ```
 
+Make sure your editor is configured to read the [`.editorconfig`](.editorconfig)
+file. This file is used to specify code formatting rules your editor. Most 
+editors
+[support it natively or have plugin that does](http://editorconfig.org/#download).
+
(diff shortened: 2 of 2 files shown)


g-bauer:
Could elaborate a bit on what this config does? Do I have to use it somehow with my editor?

Luthaf:
I don't know what to think about this. On one hand, I like the idea of editorconfig files, and having a style enforced automatically for all editors, but on the other hand, as @g-bauer noticed, everyone needs to make sure that his editor is configured to use these files, which does not resolve completely the issue, since there is still an human action that needs to be documented somewhere.

How about we both document code style somewhere, and have a lint run by Travis checking whitespaces?

---

I also tried rustfmt, and I do not like the output it gives. I know it is configurable using a `rustfmt.toml` file, but I could not manage to create something nice. My main issues where about argument placement in long function call/function definition; and the handling of small struct literals.

---

> Could elaborate a bit on what this config does? Do I have to use it somehow with my editor?

It is a file to share code style configuration in git repository that will be editted by muyltuiple people using various editors. Instead of having `.vimrc`, `.emacsrc`, `.vscoderc` and so on, there is one file that should be used by all the editors (most of the time with an additional plugin for your editor). See http://editorconfig.org about this.

antoinewdg:
I get your concerns, here is how I would do:
* keep the `.editorconfig`
* add a section in the contribution tutorial to explain what it is, and ask to use it
* I agree that a lint for whitespaces on Travis would be nice, in the near future I do not see the need for something like `rustfmt`, right now the code formatting is nice so I would not invest time in it until it becomes an issue

Luthaf:
Why not. @g-bauer, what do you think?

g-bauer:
It's fine for me. I tend to forget to add newlines at the end of the files 👍

PR #92, opened 2017-01-23
Do not divide by zero when the cell in infinite

### LINKED ISSUE #89: Crashing examples
In addition to what I mentioned in #88, the `minimization` example panics:
```
thread 'main' panicked at 'Potential energy is infinite!', /home/antoinewdg/Development/lumol/src/core/src/sys/compute.rs:119
```

I also get warnings on `mc_npt_spce` and `mc_npt_ethane`, is this intentional ?
```
Warning: No potential defined for the pair (H, O) (from /home/antoinewdg/Development/lumol/src/core/src/sys/interactions.rs:181)
Warning: No potential defined for the pair (H, H) (from /home/antoinewdg/Development/lumol/src/core/src/sys/interactions.rs:181)
Initial energy: -76663.13161568451
```
```
Warning: Tried to increase the maximum amplitude for translations to more than the set maximum. (from /home/antoinewdg/Development/lumol/src/core/src/sim/mc/moves/translate.rs:140)
Warning: Tried to increase the maximum amplitude for rotations to more than 180°. (from /home/antoinewdg/Development/lumol/src/core/src/sim/mc/moves/rotate.rs:125)
```

Luthaf:
Fix #89.

The first commit changes should not be controversial, as we assert that there are not reciprocal vectors and that tail correction does not make sense in infinite cells.

The second one is more open to discussion. When computing pressure/stress/virial, I set the volume to 1 and carry on with the computation.

TODO: 
- [x] Decide what to do for pressure/virial/stress
- [x] Add tests for pressure/virial/stress with infinite cells

### DIFF
--- a/src/core/src/sys/cells.rs
+++ b/src/core/src/sys/cells.rs
@@ -107,6 +107,12 @@ impl UnitCell {
         self.shape
     }
 
+    /// Check if this unit cell is infinite, *i.e.* if it does not have periodic boundary
+    /// conditions.
+    pub fn is_infinite(&self) -> bool {
+        self.shape() == CellShape::Infinite
+    }
+
     /// Get the first vector of the cell
     pub fn vect_a(&self) -> Vector3D {
         let x = self.cell[(0, 0)];
@@ -210,7 +216,7 @@ impl UnitCell {
         }
     }
 
-    /// Get the volume angle of the cell
+    /// Get the volume of the cell
     pub fn volume(&self) -> f64 {
         let volume = match self.shape {
             CellShape::Infinite => 0.0,
@@ -242,6 +248,9 @@ impl UnitCell {
 
     /// Get the reciprocal vectors of this unit cell
     pub fn reciprocal_vectors(&self) -> (Vector3D, Vector3D, Vector3D) {
+        assert!(!self.is_infinite(),
+            "Can not get reciprocal vectors of infinite cells"
+        );
         let volume = self.volume();
         let (a, b, c) = (self.vect_a(), self.vect_b(), self.vect_c());
 
@@ -437,6 +446,7 @@ mod tests {
     fn infinite() {
         let cell = UnitCell::new();
         assert_eq!(cell.shape(), CellShape::Infinite);
+        assert!(cell.is_infinite());
 
         assert_eq!(cell.vect_a(), Vector3D::zero());
         assert_eq!(cell.vect_b(), Vector3D::zero());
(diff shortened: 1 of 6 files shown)


src/core/src/sys/compute.rs
```diff
@@ -277,8 +286,11 @@ impl Compute for Stress {
             kinetic += particle.mass * velocity.tensorial(velocity);
         }
 
-        let virial = Virial.compute(system);
-        let volume = Volume.compute(system);
+        let volume = match system.cell().shape() {
```
antoinewdg: This piece of code appears three times, maybe it would be useful to put it in a method ?

Luthaf:
> When computing pressure/stress/virial, I set the volume to 1 and carry on with the computation.

What do you think about this? Computing pressure/stress in infinite cell is likely to be wrong, but might still be interesting. If we settle on this behaviour, we should document it.

src/core/src/sys/cells.rs
```diff
@@ -227,6 +227,18 @@ impl UnitCell {
         return volume;
     }
 
+    /// Get the volume of the unit cell if it has periodic boundary conditions, or the `default`
+    /// value.
+    ///
+    /// This function is intended for use where one need to divide unconditionally by the cell
```
g-bauer: s/need/needs

g-bauer:
Yes, there needs to be documentation for this behavior - maybe we should also display a warning (`warn_once`). 

Alternatively, we could deny those outputs that "don't make sense" for infinite cells - as you said, they are likely to be wrong or simply not defined.

Luthaf:
> Alternatively, we could deny those outputs that "don't make sense" for infinite cells - as you said, they are likely to be wrong or simply not defined.

I picked this behaviour, as it is less surprising. So now the code will panic if the user try to compute pressure/stress/virial for an infinite cell.

src/input/src/simulations/system.rs
```diff
@@ -50,7 +50,7 @@ impl Input {
         try!(self.read_potentials(&mut system));
         try!(self.init_velocities(&mut system));
 
-        if !with_cell && system.cell().shape() == CellShape::Infinite {
+        if !with_cell && system.cell().is_infinite() {
             warn!("No unit cell in the system, using an infinite unit cell.\n\
             You can silent this warning by using `cell = []` in the input file \
```
g-bauer: silent -> mute?

Luthaf:
@g-bauer or @antoinewdg, this is OK for my side !

antoinewdg:
OK for me.

PR #94, opened 2017-01-24
Input file support for `target_acceptance` for Monte-Carlo moves

g-bauer:
This PR adds input file support for adding MC moves with a target acceptance (see #23). Also, the `update_amplitude_frequency` - which is needed to achieve the acceptance - can be set from the input.

- [x] Extract unsigned integer from TOML
- [x] Add key for `target_acceptance`
- [x] Add key for ~`update_delta_every` (since all our displacements are called `delta`)~ `update_frequency`
- [x] Add example input file
- [x] Add tests in input/tests
- [x] Add documentation

Open questions:
- [x] How to handle the case, where an acceptance is given, but no update_frequency? Set default? Error?
    * Solution: Throw error telling the user to specify `update_frequency`.
- [x] How to handle the case, where an update_frequency is given, but no acceptance? Ignore? Set default for acceptance?
    * Solution: Accept input. Updates don't change scaling factor.

I'd be happy to discuss naming of keys and any other kind of feedback.


src/input/src/simulations/mc.rs
```diff
@@ -20,6 +20,12 @@ impl FromTomlWithData for MonteCarlo {
 
         let mut mc = MonteCarlo::new(temperature);
 
+        if config.get("update_delta_every").is_some() {
+            let update_amplitude = try!(extract::uint(
+                    "update_delta_every", config, "Monte-Carlo propagator"));
+            mc.set_amplitude_update_frequency(update_amplitude);
```
Luthaf: I would use the same name in input and code. Either `update_delta_every` or `update_frequency` (I prefer the second), but the same one.

This help users of the code to dive in and write rust code: they are already familiar with the naming of functions.

doc/src/input/mc.md
```diff
@@ -24,32 +24,42 @@ center of mass.
...
 
+You can add all necessary information after the `[simulations.propagator]` label.
+
+- Needed keys:
+    * `type = "MonteCarlo"`
+    * `temperature` (string): System temperature. The string contains the 
+temperature with unit.
+
+- Optional keys:
+    * `update_delta_every` (uint): After this number of steps of a move, `delta` values
```
Luthaf: `unsigned integer` or `positive integer` or just `integer` might be more readable than `uint` here.

examples/ethane.toml
```diff
@@ -0,0 +1,41 @@
+# run via:
+# cargo run ethane.toml --release
```
Luthaf: Should we merge this with the mc_npt_ethane.rs example ?

Luthaf:
> How to handle the case, where an acceptance is given, but no update_frequency? Set default? Error?
> How to handle the case, where an update_frequency is given, but no acceptance? Ignore? Set default for acceptance?

I would go for the conservative way: accept `update_frequency` without `acceptance`, but not the reverse.

examples/ethane.toml
g-bauer: What do you mean by "merge"? Use this instead of the other example?

Luthaf: I believe both examples run the same simulation? If so the `.rs` file could just read this input and run the corresponding simulation.

g-bauer: Yepp, they both run the same simulation. I thought it would be good to show how to do the exact same thing using both methods.

We can change the `.rs` though if that makes more sense.

g-bauer: That's true. I think having a lot of examples is a good thing for now. We can sort out examples showing features later.

doc/src/input/mc.md
```diff
@@ -21,32 +21,42 @@ center of mass.
...
 [simulations.propagator]
 type = "MonteCarlo"
 temperature = "500 K"
+update_frequency = 500
+
 moves = [
-    {type = "Translate", delta = "1 A", frequency = 2},
-    {type = "Rotate", delta = "20 deg", molecule = "CO2.xyz"},
-    {type = "Resize", pressure = "10 bar", delta = "3 A^3", frequency = 0.001},
+    {type = "Translate", delta = "1 A", frequency = 2, target_acceptance = 0.5},
```
Luthaf: I prefer if we do not have `target_acceptance` in the main examples, but only in the corresponding documentation. Else people will think it is mandatory and start to copy-paste it everywhere.

src/input/tests/simulation/bad/mc/move_acceptance-3.toml
```diff
@@ -0,0 +1,17 @@
...
+[[simulations]]
+nsteps = 1
+
+[simulations.propagator]
+type = "MonteCarlo"
+temperature = "300 K"
+update_frequency = -1
+#^ 'update_frequency' must be a positive integer in Monte-Carlo propagator
+moves = [
+    {type = "Rotate", delta = "6 A", frequency = 0.3, target_acceptance = 0.5}
```
Luthaf: Could you add "bad" tests setting `target_acceptance` to something below 0 and above 1?

Luthaf:
@g-bauer, could you also squash together related commits (the fix whitespace ones; the add examples ones; ...)

doc/src/input/mc.md
g-bauer: Sure, I can change that. From my experience, setting an acceptance (or having it set to 0.5 by the program) is the default behavior in most codes I know.

src/input/tests/simulation/bad/mc/move_acceptance-3.toml
g-bauer: Sure. Currently, this is not checked within `input` but in `core`. It will panic with a hint to the mcmove source file. Should we catch this in the input directly?

Luthaf: Yes, panics in core are more to check that the API is respected, but we should always respect the API from the input crate. I think it is better for the user experience to get errors instead of panics.

g-bauer:
> @g-bauer, could you also squash together related commits (the fix whitespace ones; the add examples ones; ...)

I'll try. Fingers crossed that I'm not killing my branch 😄

Luthaf:
> Edit: Aaaaaand I forgot to add the new input tests 😞

Did you? I see them in the last commit. Or you pushed a change after your comment?

g-bauer:
> Or you pushed a change after your comment?

🙄

Issue #96, opened 2017-02-05
Convert integration tests to use input files


Luthaf:
Tests in the root `tests` directory are global integration tests. They should use input files to improve readability and consistency.

We will still need some Rust code to check the system properties before/after the run, but hopefully this will be less code.

antoinewdg:
By the way, aren't examples and integration tests a little redundant ?
I don't really like the fact that we don't test our examples right now, so maybe merging integration tests and example would be a win win ?

Luthaf:
I don't think they are completely redundant: examples show how you can do things with the framework, and tests check if the code is right. We will need one example showing how to build a simulation by hand, one using the input system, and some examples showing how to extend the code (new potentials, new integrators, ...)

However, I agree that some examples are redundant and that we should do some cleanup.

> I don't really like the fact that we don't test our examples right now

The code is not run, but it is still compiled by `cargo test`, to prevent bit-rotting.

antoinewdg:
> The code is not run, but it is still compiled by cargo test, to prevent bit-rotting.

Yeah, but half the examples did not work when I tried them even if they compiled. This may not be of critical right now, but when I try out a new lib examples are one of the first things I check out, and if they don't work it's most likely a no-go for me. Maybe just adding a test that runs the example (we would have to set a reasonable number of iterations) would be enough.

I suggest to first get to work on rewriting the tests so that they use input files (I'm starting right now), then we'll see about examples.

Luthaf:
> I suggest to first get to work on rewriting the tests so that they use input files (I'm starting right now)

Nice! Do not hesitate to assign yourself to the issue when you are working on something =)

antoinewdg:
What's the use of dilating the system in [the mc-helium test](https://github.com/lumol-org/lumol/blob/master/tests/mc-helium.rs#L48) ? Is there a way to do this with input files ?

Luthaf:
> What's the use of dilating the system in the mc-helium test ?

The `helium.xyz` file contains a dense helium fluid, which does not behave like a perfect gaz. We dilate it to make it a gaz and test for the perfect gaz equation.

> Is there a way to do this with input files ?

No, and I do not think it is worth it to add. We can use a separated input file with scaled dimensions, or just us a bigger cell and wait for equilibration of the positions (this may need a longer run).

antoinewdg:
@Luthaf thanks, I'll go with a new input file

antoinewdg:
[...]

PR #97, opened 2017-02-09
[WIP] Rewrite integration tests to use input files

antoinewdg:
Issue #96 .

Right now this is a first draft to see if this seems good, I'll wait for your input to go on.

I suggest we reorganize the `data` folder a little bit so that it doesn't get too messy when the number of tests grows, and I try to use consistent naming for the files.


antoinewdg:
(damn I can click the merge button now, so.. much.. power)

Luthaf:
> ` tests/data/water-ewald.toml → tests/data/interactions/mc-water-ewald.toml `

You can drop the `mc-` part of the filename, as this will be used for both mc and md.

> `tests/data/water.xyz → tests/data/trajectories/water.xyz `

`trajectories` feels wrong here: a trajectory is a sequence of configuration produced by a simulation. I would rather use `inital` or something like this.

antoinewdg:
@Luthaf sure. `initial` feels a bit vague, maybe `positions` ? (I thought about `configuration`, but that's confusing).

antoinewdg:
About [md-water](https://github.com/lumol-org/lumol/blob/master/tests/md-water.rs#L21), I feel like having the initial configuration of molecules created with code is nice because it makes it explicit that we use molecules on a regular grid, plus I would have to recreate a file for each time we want molecules on a grid.

Is this actually something useful ? Like could this be something we allow users to put in the input file ? Something like the following:
``` toml
[[systems]]
cell = 40
regular_grid = {molecule="H2O.xyz", nx=4, ny=10, nz=10}
```

I feel this would be really useful for someone who want to fiddle with Lumol without spending too much time generating files.

Luthaf:
Sorry, I though I already had answered here.

> @Luthaf sure. initial feels a bit vague, maybe positions ? (I thought about configuration, but that's confusing).

The initial configuration can contain more than positions: for example bonding information and velocities (not implemented yet, but useful for deterministic restart). `configuration` looks OK for me.

> About md-water, I feel like having the initial configuration of molecules created with code is nice because it makes it explicit that we use molecules on a regular grid, plus I would have to recreate a file for each time we want molecules on a grid.

Again, this was mainly an artefact of the origin of the code. Having a regular grid was easier to do by hand than having random positions. So you can change this without fear!

> I feel this would be really useful for someone who want to fiddle with Lumol without spending too much time generating files. 

We could add it to the code, in a separated crate (lumol-initial? lumol-generate?). But there are already quite a few tools to generate initial configurations (not all of them are easy to use though). On the top of my head, I see [packmol](http://www.ime.unicamp.br/~martinez/packmol/home.shtml) and [VMD](http://www.ks.uiuc.edu/Research/vmd/plugins/solvate/). 

If you just need some initial configurations for tests or benchmark, ask me and I'll generate them.

antoinewdg:
> If you just need some initial configurations for tests or benchmark, ask me and I'll generate them.

In that case that would be super cool if you could generate the configuration for water in `md-water` and the dilated helium in `mc-helium`. Thanks :smile: !

antoinewdg:
For the NIST test I'll have to read only a `System` from a `toml` file. [`Input::read_system`](https://github.com/lumol-org/lumol/blob/master/src/input/src/simulations/system.rs#L19)  seems like it really doesn't want to be use outside of the `Input` class, is it fine if I used it in this case anyway ?

Luthaf:
> Input::read_system seems like it really doesn't want to be use outside of the Input class, is it fine if I used it in this case anyway ?

I can see a need for it for other code users, so we could as well make it public and documented. We could do the same for [`Input::read_simulation`](https://github.com/lumol-org/lumol/blob/bbdddcb23510e53a64f3e4626417217c69c19544/src/input/src/simulations/simulations.rs#L15).

BTW, you can hit `y` when linking to code on github, this will give you a link to the commit and not master, to ensure the link is still valid in some months.

Luthaf:
> In that case that would be super cool if you could generate the configuration for water in md-water and the dilated helium in mc-helium. Thanks 😄 !

Done here: https://gist.github.com/Luthaf/dd6af8b6f0f00279497d6fc2019c7e97. I included the python script I used to generate the files, using the unreleased version 0.7 of chemfiles (the 0.6 PDB writer had bugs).

`dilated.pdb` is the dilated helium, `water.pdb` is the water use for Wolf test and `water-small.pdb` is the water used for Ewald test. I used the PDB format because it contains the unit cell and bonding informatio, so there is no need to use `cell` or `guess_bonds` in the input.

antoinewdg:
Thanks.

> BTW, you can hit y when linking to code on github, this will give you a link to the commit and not master, to ensure the link is still valid in some months.

I feel stupid asking this, but when do I press `y` ?

Luthaf:
When you are viewing a page with `github.com/<user>/<repo>/blob/` in the URL, pressing `y` change the URL  to point to the commit.

antoinewdg:
Oh... Thanks for the tip !

antoinewdg:
I have some problems: 
* the threshold for `md-water::constant_energy_ewald` is [marked as bad](https://github.com/antoinewdg/lumol/blob/0205e767d2669e99619464e6ada2839cbe2baee8/tests/md-water.rs#L32) and the test indeed fails by a small amounts, what should we do about this ?
* the tests in `nist-lj` fail for the 3rd dataset. This is most likely due to a copy-paste mistake, but I can't find it. Can somebody take a look please ?

antoinewdg:
Are the configuration files for `mc-helium` and `md-helium::perfect_gaz` supposed to be the same ?

g-bauer:
> the tests in nist-lj fail for the 3rd dataset. This is most likely due to a copy-paste mistake, but I can't find it. Can somebody take a look please?

What error do you get here? From a first glance, it all looks good ...

Luthaf:
> the threshold for md-water::constant_energy_ewald is marked as bad and the test indeed fails by a small amounts, what should we do about this ?

I'll give this one a look, energy should really be conserved here.

> Are the configuration files for mc-helium and md-helium::perfect_gaz supposed to be the same ?

No, only the potential should be the same. The `mc-helium` should use a denser system than the perfect gaz.

antoinewdg:
@g-bauer I get the following errors:
```
---- cutoff_4_lrc::nist1 stdout ----
        thread 'cutoff_4_lrc::nist1' panicked at 'assertion failed: `(left == right)` (left: `-1175.4`, right: `-4467.5`)', tests/nist-lj.rs:166
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- cutoff_3_lrc::nist1 stdout ----
        thread 'cutoff_3_lrc::nist1' panicked at 'assertion failed: `(left == right)` (left: `-1146.7`, right: `-4351.5`)', tests/nist-lj.rs:66

```

@Luthaf could you give me the file for `md-helium::perfect_gaz`  the please ?

tests/data/nist/lj-1.toml
```diff
@@ -0,0 +1,6 @@
+[input]
+version = 1
+
+[[systems]]
+file = "configuration/lj-3.xyz"
```
g-bauer: Typo here, should be `"configuration/lj-1.xyz"`

Luthaf:
> @Luthaf could you give me the file for md-helium::perfect_gaz the please ?

This is the `dilated.pdb` file in the gist. The standard `mc-helium` test should still use the `hekium.xyz` file.

antoinewdg:
@Luthaf now I'm confused. `mc-helium` [also dilates](https://github.com/antoinewdg/lumol/blob/dbe819f52142470b94bce84472ff768fb8da5374/tests/mc-helium.rs#L48) the system.

tests/data/nist/lj-1.toml
antoinewdg: Thanks ! My bad I don't know why I thought this was the third dataset, it's clear in the errors that it's the first.

Luthaf:
> @Luthaf now I'm confused. mc-helium also dilates the system.

Wait, I though there was a separated test for mc-helium, but it looks like there is no. Which one were you thinking about?

antoinewdg:
@Luthaf there is `tests/md-helium` and `tests/mc-helium`, is that not what you mean by separated test ?

antoinewdg:
On a completely unrelated note, here is a list of problems I would like to hear your thoughts about:

- [ ] we have inconsistent name formating for files : all the files name are `separated-with-dashes` in `tests`, most are `snake_cased` in `examples`, all source files are `snake_cased` except in `scripts`
- [ ] we need to use `assert_approx_eq` everywhere it makes sense to be used, because using something like `f64::abs(pressure * volume - n * K_BOLTZMANN * temperature) < 1e-3` produces unclear error messages. Right now the macro is only defined in `core` and is only used there.
- [ ] we could even have a `assert_error_smaller` for cases like `f64::abs((config.system.pressure() - pressure)/pressure) < 5e-2`

Luthaf:
For the second and third point, we could use the [approx](https://crates.io/crates/approx) crate, I believe the current macro is equivalent to `assert_ulps_eq!`, and the macro for your third point is `assert_relative_eq!` (I still have to double check this, floating point is hard ...)

antoinewdg:
@Luthaf this is in fact not exactly the same, the source code links to [this blog post](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/) that explains floating point comparisons, this is quite an interesting read.

I'll try to integrate the crate in a separate PR to see if this could be what we want.

antoinewdg:
@Luthaf I'm still completely lost about the issue with `mc-helium` being dilated but not in the same way than `md-helium`, could you take a look please ?

Luthaf:
> @Luthaf I'm still completely lost about the issue with mc-helium being dilated but not in the same way than md-helium, could you take a look please ?

You mean the `positions x 10` vs `positions x 36` ? I just ran the test, you can use the dilated.pdb file for both tests. I think the difference is just an historical artefact.

antoinewdg:
OK !

antoinewdg:
* there's a stray `argon.xyz` in the `tests/data` folder, it's not used anywhere, can I delete it or is there a reason for its existence ?
* can I reuse the `water_ewald` interactions for the `nist_spce` tests
* if not, the `alpha` parameter of ewald is manually set in the current code, which can not be done with the input file, is there a way around it ?
* can guess_bonds be used instead of the manual setting of bonds that is currently done in the code ? of not can someone generate the .pdb files with the bounds ?

Luthaf:
> there's a stray argon.xyz in the tests/data folder, it's not used anywhere, can I delete it or is there a reason for its existence ?

It can be removed!

> can I reuse the water_ewald interactions for the nist_spce tests ?

I don't think so, because we need to set the alpha value to compare with the NIST reference data. The function to set alpha is not exported to the input because it is a very advanced setting, but we could actually add it, with a caution word in the doc. I'll try to do this.

> can guess_bonds be used instead of the manual setting of bonds that is currently done in the code ? of not can someone generate the .pdb files with the bounds ?

So I tried, and guess_bonds guess too many bonds (209 instead of 200 for spce-1, 413 instead of 400 for spce-2, 639 instead of 600 for spce-3 and 1562 instead of 1500 for spce-2). Generating PDB files could be an option, but the PDB support less digits for the positions than there are in the files, and this may lead to differences in the results. Maybe we could use [Tinker XYZ](http://chembytes.wikidot.com/tnk-tut00#toc2) for this, but the format is not yet available in chemfiles.

So I guess this test can stay as it is for now, and we can still improve it later.

antoinewdg:
Well in this case I think the PR is ready for review.

tests/data/md-nacl-npt-wolf-small.toml
```diff
@@ -14,3 +14,4 @@ nsteps = 10_000
 type = "MolecularDynamics"
 timestep = "1 fs"
 integrator = {type = "AnisoBerendsenBarostat", pressure = "5e4 bar", timestep = 1000}
+thermostat = {type = "Berendsen", temperature = "273 K", timestep = 1000}
```
Luthaf: The thermostat timestep should be smaller than the barostat timestep. A value of 100 is nice.

tests/data/md_helium/berendsen_barostat.toml
```diff
@@ -0,0 +1,20 @@
...
+lj = {sigma = "2 A", epsilon = "0.2 kJ/mol"}
+cutoff = "10 A"
+
+[[simulations]]
+nsteps = 1_000
+
+[simulations.propagator]
+type = "MolecularDynamics"
+timestep = "1 fs"
+integrator = {type = "BerendsenBarostat", pressure = "5000 bar", timestep = 1000}
```
Luthaf: This simulation need a thermostat too, the same as in the other comment should do.

tests/data/md_helium/table_computation.toml
```diff
@@ -0,0 +1,21 @@
...
+[[systems]]
+file = "../configuration/helium.xyz"
+cell = 10
+velocities = {init = "300 K"}
+
+[[systems.potentials.pairs]]
+atoms = ["He", "He"]
+lj = {sigma = "2 A", epsilon = "0.2 kJ/mol"}
+cutoff = "7 A"
+computation = {table = {max = "7 A", n = 1000}}
```
Luthaf: Could we use three separated files for the interactions here? `cutoff`, `shifted` and `table`?

tests/data/md-nacl-npt-wolf-small.toml
Luthaf: The thermostat timestep should be smaller than the barostat timestep. A value of 100 should be good.

Luthaf:
Also, can you rebase this PR and squash "fix ..." commits with the fixed commit?

antoinewdg:
Can't we just squash everything into one big commit ? It's not like the precise history is all that interesting.

Luthaf:
Yes we can do this too. I'll do a second round a review after to double check everything locally.

antoinewdg:
Since I've done some merges before rebasing is painful, I'll just extract a patch and create a clean PR

tests/data/md_helium/berendsen_barostat.toml
antoinewdg: I don't see any other comment with a thermostat :( Is this one OK ?
``` toml
thermostat = {type = "Berendsen", temperature = "273 K", timestep = 1000}
```

tests/data/md_helium/table_computation.toml
antoinewdg: Yes probably, but I think I'm missing the point here, could you explain why you want to do that ?

tests/data/md_helium/berendsen_barostat.toml
Luthaf: I tried to add this comment bu I was not sure. So yes, this one is OK, but the timestep would be better at 100. It needs to be lower than the barostat timestep.

tests/data/md_helium/table_computation.toml
Luthaf: Just to make it more clear what is going on. It is more a suggestion than a requirement!

tests/data/md_nacl_npt_wolf_small.toml
```diff
@@ -0,0 +1,17 @@
...
+velocities = {init = "300 K"}
+
+[[simulations]]
+nsteps = 10_000
+
+[simulations.propagator]
+type = "MolecularDynamics"
+timestep = "1 fs"
+integrator = {type = "AnisoBerendsenBarostat", pressure = "5e4 bar", timestep = 1000}
+thermostat = {type = "Berendsen", temperature = "273 K", timestep = 1000}
```
Luthaf: The other comment I was talking about was this one, but it got lost on file renaming. Could you also change the thermostat timestep here to 100?

tests/data/md_helium/berendsen_barostat.toml
antoinewdg: Fixed in #116

tests/data/md_nacl_npt_wolf_small.toml
antoinewdg: Fixed in #116

tests/data/md_helium/table_computation.toml
antoinewdg: Since this would still require three different main `.toml` files I feel this would just get more complicated for no real gain.

Luthaf:
Closing in favor of #116

PR #99, opened 2017-02-13
Add Buckingham and Born-Mayer-Huggins potentials

Luthaf:
This PR add the very common Buckingham potential, and the less common Born-Mayer-Huggins potential (derived from the Buckingham potential).

Two questions for the reviewers:
    - I used the expressions from LAMMPS ([born](http://lammps.sandia.gov/doc/pair_born.html), [buckingham](http://lammps.sandia.gov/doc/pair_buck.html)), using `1/rho` instead of `B` in the exponential. [Wikipedia](https://en.wikipedia.org/wiki/Buckingham_potential) and [Sklog Wiki](http://www.sklogwiki.org/SklogWiki/index.php/Buckingham_potential) uses the other form. I prefer the one with rho because it make it clearer it is a length parameter. Is it OK for you?
    - I used the `born` keyword in input instead of the verbose `Born-Mayer-Huggins`, is this OK? I am following LAMMPS here too.

Ref #6

### DIFF
--- a/doc/src/input/potentials.md
+++ b/doc/src/input/potentials.md
@@ -47,6 +47,40 @@ atoms = ["O", "O"]
 lennardjones = {sigma = "3.16 A", epsilon = "0.155 kcal/mol"}
 ```
 
+## Buckingham potential
+
+The Buckingham potential is a potential for pair interactions expressed as: $$
+V(x) = A \exp(-r / \rho) - \frac{C}{r^6}.$$
+
+The potential type keyword is `buckingham`, and the parameters `A`, `rho`
+($\rho$) and `C` should be provided as strings.
+
+```toml
+[[pairs]]
+atoms = ["A", "B"]
+buckingham = {A = "40 kJ/mol", C = "120e-6 kJ/mol/A^6", rho = "3.0 A"}
+```
+
+## Born-Mayer-Huggins potential
+
+The Born-Mayer-Huggins potential is a potential for pair interactions, used in
+particular for halide alkali. Its expression is: $$ V(x) = A
+\exp\left(\frac{\sigma -r}{\rho}\right) - \frac{C}{r^6} + \frac{D}{r^8}.$$
+
+The potential type keyword is `born`, and the parameters `A`, `C`, `D`, `sigma`
+($\sigma$) and `rho` ($\rho$) should be provided as strings.
+
+```toml
+[[pairs]]
+atoms = ["A", "B"]
+[pairs.born]
+A = "40 kJ/mol"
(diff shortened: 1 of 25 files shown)


g-bauer:
> I used the expressions from LAMMPS (born, buckingham), using 1/rho instead of B in the exponential. Wikipedia and Sklog Wiki uses the other form. I prefer the one with rho because it make it clearer it is a length parameter. Is it OK for you?

I only knew the expression used in Sklog Wiki but I think the other expression is fine, too, and there should be no confusion since you did a clear documentation.

> I used the born keyword in input instead of the verbose Born-Mayer-Huggins, is this OK? I am following LAMMPS here too.

Fine for me. In case of doubt, we can do something similar like the Lennard Jones case, where we have both `lj` and `lennardjones` as keys?

I'll have time tonight to review the sources.

Luthaf:
> Fine for me. In case of doubt, we can do something similar like the Lennard Jones case, where we have both lj and lennardjones as keys?

I think Lennard-Jones potential is common enough that we can use `lj` for it, and forbid `lennardjones`. It might be better to have only one way of doing things.

doc/src/input/potentials.md
```diff
@@ -47,6 +47,39 @@ atoms = ["O", "O"]
...
+
+```toml
+[[pairs]]
+atoms = ["A", "B"]
+buckingham = {A = "40 kJ/mol", C = "120e-6 kJ/mol/A^6", rho = "3.0 A"}
+```
+
+## Born-Mayer-Huggins potential
+
+The Born-Mayer-Huggins potential is a potential for pair interactions, used in particular for halide alkali. It expression is: $$
```
g-bauer: s/It/Its

src/core/src/energy/functions.rs
```diff
@@ -247,6 +247,134 @@ impl Potential for Torsion {
 
 impl DihedralPotential for Torsion {}
 
+/// Buckingham potential.
+///
+/// The following potential expression is used: `V(x) = A * exp((sigma - r) /
+/// rho) - C/r^6`; where `A` and `C` are energetic constants, and `rho` is a
+/// length parameters.
```
g-bauer: s/parameters/parameter

g-bauer:
Very nice. I'll merge as soon as you give the go. Btw. is it ok for the reviewer to change small typos and like the above before merging?

Luthaf:
> Btw. is it ok for the reviewer to change small typos and like the above before merging?

Yes, definitively! I've done it for this one, but feel free for the next ones =)

PR #91, opened 2017-01-23
Check for whitespaces on CI

Luthaf:
Follow-up of #90, adding checks for whitespaces on Travis.

The tests should fail, I will fix all of them when I am sure this makes Travis fail.

PR #101, opened 2017-02-20
Use cargo test --all in travis and in doc

Luthaf:
Support for `cargo test --all` landed in rust 1.15. We can use it for release tests, but debug tests are very slow. instead, we can specify which package to test with `cargo test -p <package>`. It feels more convenient that changing directory all the time, so let's use it in Travis and in contributors documentation.

TODO before merging:
- [x] check that coverage is still OK


antoinewdg:
It's nice that you fixed the weird warning we discussed at the beginning of #81 .
By the way why has Codecov been disabled ?

Luthaf:
> By the way why has Codecov been disabled ?

No idea ... Looks like the last build that hit codecov was [20 days ago](https://codecov.io/gh/lumol-org/lumol/branches). I'll give it a look tonight.

Luthaf:
Code coverage is working again \o/

antoinewdg:
So what happened ?

Luthaf:
I think I broke it somehow in #91, and the name of the test executables changed with this PR too.

The coverage generation is no longer part of the `after_script` section, but of the `script` section, which means that failing coverage will fail the test, and will not stay unnoticed.

Luthaf:
Do we want to remove the codecov/patch check? It seems it gives a lot of false negatives. Or maybe we can use it only if there is more than 100 lines changed (with a suitable line count threshold).

Also, the code coverage went up again by 3% (yay \o/), I have no idea why (=/). I am explicitly [linking dead code](https://github.com/lumol-org/lumol/blob/4fb238d3d8f745e5de8dfae31a25678e2d9c487f/.travis.yml#L46) here.

Do we want to modify the threshold for codecov/project then?

PR #102, opened 2017-02-20
Only accept integers as 'every' in controls

### LINKED ISSUE #31: Ability to run control algorithm not at all steps

Luthaf:
Ref #84 and fix #31 !

### DIFF
--- a/src/input/src/simulations/md.rs
+++ b/src/input/src/simulations/md.rs
@@ -172,7 +172,7 @@ impl FromToml for RescaleThermostat {
 impl FromToml for Alternator<RemoveTranslation> {
     fn from_toml(config: &Table) -> Result<Alternator<RemoveTranslation>> {
         let every = if config.contains_key("every") {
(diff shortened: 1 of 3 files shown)


antoinewdg:
Seems good to me.

Luthaf:
> Seems good to me.

You can hit the merge button then! 😉

antoinewdg:
I'm almost sure the answer is yes, but better safe than sorry: do I delete the branch ?

g-bauer:
Yepp!

Luthaf:
> Yepp! Oh wait ... I let the PR initiator delete it in the past. What's the right thing to do?

I don't really care here =) If everyone is OK with it we can settle on using one branch for one PR only, and delete them when merging.

Also, when creating branches on this repo (lumol-org/lumol), we might want to use some kind of namespacing scheme (i.e. this branch should be named `luthaf/every`) to prevent conflicts.

Issue #107, opened 2017-02-26
Tests on lumol core failing on my machine


antoinewdg:
All is in the title. As of commit cd2e2c0b I have the followign failure on the core tests:

``` bash
failures:

---- sys::cache::tests::move_atoms stdout ----
        thread 'sys::cache::tests::move_atoms' panicked at 'assertion failed: `(left ≈ right)` (left: `14808.992701116425`, right: `14808.992701116422`) with ∆=3.6e-12 (allowed ∆=1e-14)', src/sys/cache.rs:458

```

It stops failing if we replace it by the `assert_ulps_eq!` macro from [approx](http://brendanzab.github.io/approx/approx/index.html).

Luthaf:
Can you check #108 fixed this? It did on my machine, but better be sure than sorry !

antoinewdg:
It does :+1: .
But this is a reminder that choosing the epsilon values on the float tests is not easy !

PR #108, opened 2017-02-26
Integrate the approx crate

antoinewdg:
What this PR does:
* add the [approx](https://crates.io/crates/approx) crate to dependencies
* implement the `ApproxEq` trait for `Vector3D`
* changes some assertions in existing tests on Vector3D
* changes the problematic assertions mentionned in #107 

He are the guidelines I suggest for floating point comparison (feedback appreciated): 
* when we want two floats (are structs involving floats such as `Vector3D`) to be exactly equal in theory but they are not due two float imprecision, we use `assert_ulps_eq!` with default parameters
* when we want two floats to be close but expect them to be a little different (such as when testing that energy remains constant) we use `assert_relative_eq!` and play with the `max_relative` parameter to allow more or less error
* we never ever use `assert_eq!` for anything that has a float in it


Luthaf:
> He are the guidelines I suggest for floating point comparison (feedback appreciated):

These seems sensible guidelines for me. Can we have some kind of lint/tool to enforce "we never ever use assert_eq! for anything that has a float in it"?


Can we remove the `assert_approx_eq` macro completely and replace it by the `assert_ulps_eq`  everywhere?

antoinewdg:
> Can we remove the assert_approx_eq macro completely and replace it by the assert_ulps_eq everywhere?

We could, but what we would actually need to do is look at every single use of `assert_approx_eq` and think about if we want to use `assert_ulps_eq` or `assert_relative_eq`. I'm looking into doing it, this may be a bit long.

> Can we have some kind of lint/tool to enforce "we never ever use assert_eq! for anything that has a float in it"?

I have no idea about that, but we would have to be sure that we fix the old code before setting this up.

It turns out that having to type the `max_relative` parameter for the `assert_relative_eq!` is a bit too verbose, I suggest we create a new macro `assert_percentage_eq` that just takes the percentage as a parameter (it could also take a real percentage as a parameter instead of a number between 0 and 1), what do you think ?

Luthaf:
> We could, but what we would actually need to do is look at every single use of assert_approx_eq and think about if we want to use assert_ulps_eq or assert_relative_eq. 

Can you have a look and do the one you know in which category they fall, and I'll do the remaining ones.

> I suggest we create a new macro assert_percentage_eq that just takes the percentage as a parameter (it could also take a real percentage as a parameter instead of a number between 0 and 1), what do you think ?

Yes, why not.

Luthaf:
I found the time to finish this, now `assert_approx_eq!` is gone ! I'll merge it tomorrow if nobody has an objection.

PR #113, opened 2017-03-06
Update license to BSD-3-clauses

Luthaf:
The third clause is actually nice, especially in the context of scientific software.

Please check the box bellow if you agree:
- [x] @g-bauer 
- [x] @antoinewdg

I made this commits after running [cargo-lichking](https://github.com/Nemo157/cargo-lichking) on this repository. I would like to run the test on Travis, but it is currently failling because chemfiles is MPL. From my understanding, using MPL code inside other code is OK as long as the code is not modified. But this might be another argument to switch chemfiles to BSD (https://github.com/chemfiles/chemfiles/issues/48)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Luthaf <luthaf@luthaf.fr>"]
 documentation = "https://lumol-org.github.io/lumol"
 repository = "https://github.com/lumol-org/lumol"
 readme = "README.md"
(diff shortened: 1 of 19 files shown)


Luthaf:
Oh, I broke the coverage. Guess it is time to improve it a bit.

antoinewdg:
OK on my side.

g-bauer:
Fine for me.

PR #116, opened 2017-03-12
Rewrite integration tests

antoinewdg:
This PR is a rewrite of the integration tests located in `tests/`, so that they use input files instead of generating systems with code.

Due to the fact that the `alpha` parameter of `Ewald` energy can currently not be set in input files, `tests/nist_spce` is left as is.


Luthaf:
The issue with the NPT simulations was that we checked the last instantaneous value instead of the average over the end of the simulation. This should be fixed.

Also, the md_helium test needed a simulation a bit longer: the barostat/thermostat timestep is an exponential factor with Berendsen, so we need a simulation that last at least 5 time this factor to get close enough to the target value.

Else, everything looks good to me. I am (for no rational reason) slightly bothered by the use of underscores instead of dashes in the file names (`mc_helium` instead of `mc-helium`). Do you feel strongly about it, or can we change the normalization to dashes? Either way, I'll accustom myself to this!

antoinewdg:
I used underscore because this is the convention I saw for rust files almost everywhere else. If you really like dashes we can use them, but in this case I'm not clear how we would decide where to use dashes and where to use underscores: do only test files take dashes ?

Luthaf:
I prefer underscores for variables names, and no dashes nor underscores for modules names.

Then yes, only files in tests get dashes ! It is also better to have `md-helium.rs` and `mc-helium.rs` instead of modules (`md/mod.rs`; `md/helium.rs`; `mc/mod.rs`, `mc/helium.rs`) for integration tests because then cargo automatically recognise test cases on a file by file basis.

Luthaf:
If you are OK with this, I can do the changes and then get this merged.

antoinewdg:
I'm OK with this.

Luthaf:
So I went a bit ahead and changed the data files layout to have one folder by tests in `test/data/<case>`, containing all the files used by the corresponding `<case>.rs` file.

I believe this will make navigating the tests data files easier that trying to follow the links and having to read the code.

If everything is good for you, we can merge this. And thanks for tackling this work, it is not the most interesting part but it needed to be done.

Fix #96

antoinewdg:
This new layout is much better !

PR #121, opened 2017-03-15
Improve performance on the Ewald forces

antoinewdg:
This PR introduces the following improvements:
1. in [`Ewald::kspace_force_factor`](https://github.com/lumol-org/lumol/blob/592e171cdbcc34f043f5653b347f97bfdf4e04c0/src/core/src/energy/global/ewald.rs#L416-L422) we perform some useless computations because we use only the imaginary part of a product between complex numbers, I introduced a method in the `Complex` class to change that.
2. I took the `expfactor` out of `Ewald::kspace_force_factor` because it is independent of `i` and `j`
3. I removed some redundant multplications
4. I moved the computation of `fourier_i` because it is independent of `j` 

Point 4. is the big hitter, it gives a 30% speedup. Point 2. also seems to have a big influence (but far less noticeable).

### DIFF
--- a/src/core/src/energy/global/ewald.rs
+++ b/src/core/src/energy/global/ewald.rs
@@ -398,13 +398,16 @@ impl Ewald {
                     if expfactor < f64::EPSILON {continue}
 
                     let f = expfactor * factor;
-
                     let k = (ikx as f64) * rec_kx + (iky as f64) * rec_ky + (ikz as f64) * rec_kz;
+
                     for i in 0..system.size() {
                         let qi = system[i].charge;
+                        let fourier_i = self.fourier_phases[(ikx, i, 0)].imag_mul(
+                            self.fourier_phases[(iky, i, 1)] * self.fourier_phases[(ikz, i, 2)]
+                        );
                         for j in (i + 1)..system.size() {
                             let qj = system[j].charge;
-                            let force = f * self.kspace_force_factor(i, j, ikx, iky, ikz, qi, qj) * k;
+                            let force = f * self.kspace_force_factor(j, ikx, iky, ikz, qi, qj, fourier_i) * k;
                             callback(i, j, force);
                         }
                     }
@@ -427,10 +430,9 @@ impl Ewald {
     /// `qj`, at k point  `(ikx, iky, ikz)`
     #[inline]
     #[allow(too_many_arguments)]
-    fn kspace_force_factor(&self, i: usize, j: usize, ikx: usize, iky: usize, ikz: usize, qi: f64, qj: f64) -> f64 {
-        let fourier_i = self.fourier_phases[(ikx, i, 0)].imag_mul(
-            self.fourier_phases[(iky, i, 1)] * self.fourier_phases[(ikz, i, 2)]
-        );
+    fn kspace_force_factor(&self, j: usize, ikx: usize, iky: usize, ikz: usize,
+                           qi: f64, qj: f64, fourier_i: f64) -> f64 {
+
         let fourier_j = self.fourier_phases[(ikx, j, 0)].imag_mul(
             self.fourier_phases[(iky, j, 1)] * self.fourier_phases[(ikz, j, 2)]
         );
--- a/src/core/src/energy/global/ewald.rs
+++ b/src/core/src/energy/global/ewald.rs
@@ -378,12 +378,9 @@ impl Ewald {
         return energy;
     }
 
-    /// Loop over each k_space force.
-    ///
-    /// Abstracts over the very similar computations in `kspace_forces`
-    /// and `kspace_virial`.
-    fn for_each_kspace_force<F>(&mut self, system: &System, mut callback: F)
-    where F : FnMut(usize, usize, Vector3D){
+    /// k-space contribution to the forces
+    fn kspace_forces(&mut self, system: &System, forces: &mut [Vector3D]) {
+        assert_eq!(forces.len(), system.size());
         self.density_fft(system);
 
         let factor = 4.0 * PI / (system.cell().volume() * ELCC);
@@ -394,36 +391,34 @@ impl Ewald {
                 for ikz in 0..self.kmax {
                     // The k = 0 and the cutoff in k-space are already handled
                     // in `expfactors`.
-                    let expfactor = self.expfactors[(ikx, iky, ikz)].abs() ;
-                    if expfactor < f64::EPSILON {continue}
+                    let expfactor = self.expfactors[(ikx, iky, ikz)].abs();
+                    if expfactor < f64::EPSILON { continue }
 
                     let f = expfactor * factor;
                     let k = (ikx as f64) * rec_kx + (iky as f64) * rec_ky + (ikz as f64) * rec_kz;
 
                     for i in 0..system.size() {
                         let qi = system[i].charge;
-                        let fourier_i = self.fourier_phases[(ikx, i, 0)].imag_mul(
-                            self.fourier_phases[(iky, i, 1)] * self.fourier_phases[(ikz, i, 2)]
-                        );
+
+                        let fourier_i = self.fourier_phases[(ikx, i, 0)] *
+                                        self.fourier_phases[(iky, i, 1)] *
+                                        self.fourier_phases[(ikz, i, 2)];
+                        let fourier_i = fourier_i.imag();
+
+                        let mut force_i = Vector3D::zero();
+
                         for j in (i + 1)..system.size() {
                             let qj = system[j].charge;
                             let force = f * self.kspace_force_factor(j, ikx, iky, ikz, qi, qj, fourier_i) * k;
-                            callback(i, j, force);
+                            force_i -= force;
+                            forces[j] += force;
                         }
+
+                        forces[i] += force_i;
                     }
                 }
             }
         }
-    }
-
-    /// k-space contribution to the forces
-    fn kspace_forces(&mut self, system: &System, forces: &mut [Vector3D]) {
-        assert_eq!(forces.len(), system.size());
-
-        self.for_each_kspace_force(system, |i, j, force|{
-            forces[i] -= force;
-            forces[j] += force;
-        });
     }
 
     /// Get the force factor for particles `i` and `j` with charges `qi` and
@@ -433,23 +428,56 @@ impl Ewald {
     fn kspace_force_factor(&self, j: usize, ikx: usize, iky: usize, ikz: usize,
                            qi: f64, qj: f64, fourier_i: f64) -> f64 {
 
-        let fourier_j = self.fourier_phases[(ikx, j, 0)].imag_mul(
-            self.fourier_phases[(iky, j, 1)] * self.fourier_phases[(ikz, j, 2)]
-        );
+        // Here the compiler is smart enough to optimize away
+        // the useless computation of the last real part.
+        let fourier_j = self.fourier_phases[(ikx, j, 0)] *
+                        self.fourier_phases[(iky, j, 1)] *
+                        self.fourier_phases[(ikz, j, 2)];
+        let fourier_j = fourier_j.imag();
 
         return qi * qj * (fourier_i - fourier_j);
     }
 
     /// k-space contribution to the virial
     fn kspace_virial(&mut self, system: &System) -> Matrix3 {
+        self.density_fft(system);
         let mut virial = Matrix3::zero();
 
-        self.for_each_kspace_force(system, |i, j, force|{
-            let rij = system.nearest_image(i, j);
-            virial += force.tensorial(&rij);
-        });
+        let factor = 4.0 * PI / (system.cell().volume() * ELCC);
+        let (rec_kx, rec_ky, rec_kz) = system.cell().reciprocal_vectors();
 
-        return virial;
+        for ikx in 0..self.kmax {
+            for iky in 0..self.kmax {
+                for ikz in 0..self.kmax {
+                    // The k = 0 and the cutoff in k-space are already handled
+                    // in `expfactors`.
+                    let expfactor = self.expfactors[(ikx, iky, ikz)].abs();
+                    if expfactor < f64::EPSILON { continue }
+
+                    let f = expfactor * factor;
+                    let k = (ikx as f64) * rec_kx + (iky as f64) * rec_ky + (ikz as f64) * rec_kz;
+
+                    for i in 0..system.size() {
+                        let qi = system[i].charge;
+
+                        let fourier_i = self.fourier_phases[(ikx, i, 0)] *
+                                        self.fourier_phases[(iky, i, 1)] *
+                                        self.fourier_phases[(ikz, i, 2)];
+                        let fourier_i = fourier_i.imag();
+
+                        for j in (i + 1)..system.size() {
+                            let qj = system[j].charge;
+                            let force = f * self.kspace_force_factor(j, ikx, iky, ikz, qi, qj, fourier_i) * k;
+                            let rij = system.nearest_image(i, j);
+                            virial += force.tensorial(&rij);
+                        }
+
+                    }
+                }
+            }
+        }
+
+        virial
     }
 
     fn compute_delta_rho_move_particles(&mut self, system: &System, idxes: &[usize], newpos: &[Vector3D]) {
--- a/src/core/src/energy/global/ewald.rs
+++ b/src/core/src/energy/global/ewald.rs
@@ -378,9 +378,12 @@ impl Ewald {
         return energy;
     }
 
-    /// k-space contribution to the forces
-    fn kspace_forces(&mut self, system: &System, forces: &mut [Vector3D]) {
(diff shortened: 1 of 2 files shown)


Luthaf:
>  A macro seems the simplest way to do it, what do you think ?

I am not really fond of Rust macro_rules!, because they are hard to read. I use them if I really need to, and preferably for small pieces of code. But I agree that copy/paste is not really an enjoyable process, especially here.

Maybe we can refactor the code to share most of it? Like having a function that takes a closure:

```rust
fn do_with_forces(&self, callback: Fn(i, j, force, rij)) {
    for ikx in ... {
        for iky in ... {
            for ikz in ... {
                for i in ... {
                    for j in ... {
                        let rij = ...;
                        let force = ...;

                        callback(i, j, force, rij);
                    }
                }
            }
        }
    }
}

fn kspace_forces(&self) {
    let mut forces = vec![...];
    self.do_with_forces(|i, j, force, _|{
        forces[i] += force;
        forces[j] -= force;
    });
}

fn kspace_virial(&self) {
    let mut virial = Matrix;
    self.do_with_forces(|_, _, force, rij|{
        virial += force.tensorial(rij);
    });
}
```

antoinewdg:
> I am not really fond of Rust macro_rules!, because they are hard to read.

I 100 % agree on that.

> Maybe we can refactor the code to share most of it? Like having a function that takes a closure:

Thing is we would also need to call a closure that is called on each `i` increment because of [this](https://github.com/lumol-org/lumol/pull/121/files#diff-d4a21536e09da2159b99e9264b0cfb4aR404) line, and we may have some ownership shenanigans.
But I'm not even sure this line is that useful, so maybe I'll just forget it.

antoinewdg:
I rewrote the changes and broke them into multiple commits. Up to `f58ca8d` we are pretty good, the code duplication is removed as we wantedand we already have a significant boost in performance (30% for `Ewald::forces` and 15% for `Ewald::virial`). 

We can do better with the last commit `fce05d6` but keeping the code without duplication becomes much less readable (all of the `CallBack` related stuff is only there to enable us this last change). This is still a significant imporvement so I want to keep it, but I think if we keep go back to the duplicated code, even if I don't like it much.

antoinewdg:
[Here](https://gist.github.com/antoinewdg/0d6fb4297584ddcb01c939c61af7e21d) are the benchmarks for each commit (to avoid the warm-up issues I met I ran the benchmarks three times and kept the last one).

Surprinsigly commit `2eed55c` doesn't seem to do much good, while it obviously reduces the amount of computation. Do you think the compiler was smart enough to realize the real part was never used and optimize it away ? I find it unlikely, but I'm no compiler expert,don't know how to read assembly and have not other explanation.

Luthaf:
Great! And thanks for the benchmarks.

Concerning the imag_mul stuff, the compiler might have been clever enough to see that there is no side effect when computing the real part, and that we do not use it later, so it can just remove the computation.

I agree that the `CallBack` trait and related on the last commit is not really readable, we can have duplicated code here. I mean, performance tuning is a good reason not to stay [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

Everything else looks good to me, you might want to correct the typos in the commit messages before merging. =)

PS: looks like codecov is gone again =/

antoinewdg:
> Concerning the imag_mul stuff, the compiler might have been clever enough to see that there is no side effect when computing the real part, and that we do not use it later, so it can just remove the computation.

That's quite impressive. I prefer keeping the version where we explicitly ignore, so that the next guy looking into optimization doesn't have to ask this question.

> I agree that the CallBack trait and related on the last commit is not really readable, we can have duplicated code here. I mean, performance tuning is a good reason not to stay DRY.

I went half-measure here (I removed the previous "Uniformize memory access" commit, e94dde9 is the new one), I think it is a good trade-off. 

>  you might want to correct the typos in the commit messages before merging. =)

I think I've got them all.

I think I'm done working on this one, feel free to merge when you're OK.

Luthaf:
So I am globally OK with the code, but I still have two readability concerns (both are open to argumentation!)

```rust
let fourier_i = self.fourier_phases[(ikx, i, 0)].imag_mul(
       self.fourier_phases[(iky, i, 1)] * self.fourier_phases[(ikz, i, 2)]
);
// vs
let fourier_i = self.fourier_phases[(ikx, i, 0)] *
                self.fourier_phases[(iky, i, 1)] *
                self.fourier_phases[(ikz, i, 2)];
let fourier_i = fourier_i.imag();
```

In the second case, I can see directly what is going on. In the first (the current code), I need to look for the `fn imag_mul` definition first, because it is not a standard function. Given there is no speed difference, I would prefer to stick with the explicit version of the code. We can definitively add a comment saying that there is no need to optimize this computation because the compiler is already clever enough.

Concerning the uniform memory access, the previous version of the code with a small callback was nice and readable, but I don't like the new version as much. Now, to understand what is going on, I need to look for `kspace_forces`, then see we are calling `for_each_k_with_factor`, jump here, read the code, and then jump back to the callback inside `kspace_forces`. Given this, I think I prefer to have duplicated code that is easier to read, rather than non-duplicated code that makes me jump 3 times to get the whole computation.

antoinewdg:
About the complex product I 100 % agree I'll roll this back.

About the other point I'm not so sure. You do make good points, but I'm not convinced they outweigh the reduced code duplication. Even if the code is a little more straightforward your way, someone that reads this part will still have to invest a bit of time in learning (or recalling) what Ewald summation is, which is not exactly trivial, so having to read two different functions seems negligible in comparison.
This is a close call, and I'm definitely not dead set on this answer, if you really think this is an issue I'll just duplicate.

antoinewdg:
Actually after giving it more thought I'm agreeing with you for the duplication @Luthaf, I'll change that too.

antoinewdg:
After some benchmarking, removing the abstraction gives a slight (7%) speedup, so there's no thinking about it twice anymore :)

antoinewdg:
(there's no new commit because I squashed, but the changes are here)

Also FYI, I tried changing the array accessed in the most nested loop by unsafe unchecked accessed, and it changed absolutely nothing.

Luthaf:
This is the final result of this, comparing master to this PR on my machine:

```
$ cargo benchcmp before.log after.log --threshold 1
 name                                  before.log ns/iter  after.log ns/iter  diff ns/iter   diff %
 cache_move_all_rigid_molecules        1,052,865           974,035                 -78,830   -7.49%
 cache_move_all_rigid_molecules_ewald  8,151,369           7,570,020              -581,349   -7.13%
 cache_move_all_rigid_molecules_wolf   7,420,680           7,688,496               267,816    3.61%
 cache_move_particle                   198,749             165,185                 -33,564  -16.89%
 cache_move_particle_ewald             358,860             309,136                 -49,724  -13.86%
 cache_move_particle_wolf              92,894              87,234                   -5,660   -6.09%
 cache_move_particles                  249,794             223,780                 -26,014  -10.41%
 cache_move_particles_ewald            723,064             676,767                 -46,297   -6.40%
 cache_move_particles_wolf             455,803             448,515                  -7,288   -1.60%
 energy                                2,193,441           1,961,486              -231,955  -10.57%
 energy                                1,134,335           1,063,373               -70,962   -6.26%
 energy_ewald                          593,781             528,416                 -65,365  -11.01%
 energy_ewald                          966,866             902,253                 -64,613   -6.68%
 energy_wolf                           442,531             414,905                 -27,626   -6.24%
 energy_wolf                           690,919             658,372                 -32,547   -4.71%
 forces                                2,464,803           2,231,004              -233,799   -9.49%
 forces                                1,357,804           1,258,161               -99,643   -7.34%
 forces_ewald                          24,966,241          12,761,275          -12,204,966  -48.89%
 forces_ewald                          36,058,256          19,845,716          -16,212,540  -44.96%
 forces_wolf                           555,398             499,483                 -55,915  -10.07%
 virial                                2,865,758           2,622,869              -242,889   -8.48%
 virial                                1,432,569           1,302,021              -130,548   -9.11%
 virial_ewald                          62,676,628          50,636,727          -12,039,901  -19.21%
 virial_ewald                          92,132,224          84,119,007           -8,013,217   -8.70%
 virial_wolf                           566,905             510,833                 -56,072   -9.89%
 virial_wolf                           853,204             880,465                  27,261    3.20%
```

The forces are nearly twice as fast !

g-bauer:
Wow, good job!

antoinewdg:
@Luthaf your benchmark looks a bit skewed because everything runs about 5% faster, event the computations that were not changed.

Here is the comparison run on Travis CI (with all the cache stuff hidden):
```bash
$ cargo benchcmp old_stuff new_stuff --threshold=3
 name                                  old_stuff ns/iter  new_stuff ns/iter  diff ns/iter   diff % 
 energy_ewald                          767,671            728,044                 -39,627   -5.16% 
 forces_ewald                          27,431,955         15,705,374          -11,726,581  -42.75% 
 forces_ewald                          40,147,134         22,880,622          -17,266,512  -43.01% 
 virial_ewald                          67,403,446         57,020,094          -10,383,352  -15.40% 
 virial_ewald                          99,826,124         84,245,257          -15,580,867  -15.61% 

```

I'm a little surprised the improvement is not that good for the virial, but hey 40% on the forces is already satisfying :smile: .

Luthaf:
> Also FYI, I tried changing the array accessed in the most nested loop by unsafe unchecked accessed, and it changed absolutely nothing.

Thinking again about this, this is because we already remove the bound checks here: https://github.com/lumol-org/lumol/blob/740eee0732bd424c79c7aa3dd9f1bc7bb9cb74bd/src/core/src/types/arrays.rs#L119-L122

antoinewdg:
Wow I did not see that, shouldn't we at least name the methods `unsafe_index` or something like it to warn the developers that we're not doing the standard Rust bound checking by default ?

Luthaf:
I don't know. The point here was to still use the convenient syntax `array[(i, j, k)]` (which is already a bit cumbersome in this kind of numeric code due to the usage of tuple); not having to write `array.unsafe_get((i, j, k)) + array.unsafe_get((i, j, k + 1)) + array.unsafe_get((i, j, k + 2))`.

The best thing to do would be to remove the cost for bounds checking, but this was not negligible at the time I wrote this.

We should document it better anyway! Also, keep in mind that we are using bounds checks in debug mode here, and we can run the whole test suite in debug mode (I do it from time to time, but it is a bit too slow to run on Travis).

antoinewdg:
I get that this makes the code much more readable (and this is indeed valuable), but we are removing the unsafe marker from a function that *is* unsafe (even if it is only in production). I don't even know where we could document it, since nobody is going to read the documentation of the `Index` trait (and everyone is going to expect bound checking here).
The Rust propaganda got to me, I'm scared of reading array elements :(.

Luthaf:
> I don't even know where we could document it

We could do this in the documentation of the wrapper.

>  The Rust propaganda got to me, I'm scared of reading array elements :(.

And in C/C++/Fortran we would be arguing that adding `-fbounds-check` have a way to big overhead to be enabled by default 😄 

I think this is one of the cases where we can use unsafe in Rust: improvement of performances. I'll run a quick benchmark with and without it just to get better knowledge of the cost for bounds checking.

Luthaf:
Here are the benchmark results:

```
$ cargo benchcmp safe.log unsafe.log --threshold=2
 name                                  safe.log ns/iter  unsafe.log ns/iter  diff ns/iter   diff % 
 cache_move_all_rigid_molecules        1,463,817         1,168,495               -295,322  -20.17% 
 cache_move_all_rigid_molecules_ewald  10,620,984        8,721,128             -1,899,856  -17.89% 
 cache_move_all_rigid_molecules_wolf   8,386,407         10,278,223             1,891,816   22.56% 
 cache_move_particle                   201,875           244,669                   42,794   21.20% 
 cache_move_particle_ewald             528,745           491,710                  -37,035   -7.00% 
 cache_move_particles                  354,655           282,355                  -72,300  -20.39% 
 cache_move_particles_ewald            1,025,460         887,721                 -137,739  -13.43% 
 cache_move_particles_wolf             507,774           553,008                   45,234    8.91% 
 energy                                2,380,384         2,766,820                386,436   16.23% 
 energy_ewald                          785,943           674,102                 -111,841  -14.23% 
 energy_ewald                          1,257,053         1,126,991               -130,062  -10.35% 
 forces                                2,545,932         3,001,106                455,174   17.88% 
 forces                                1,408,668         1,459,264                 50,596    3.59% 
 forces_ewald                          18,974,768        13,906,190            -5,068,578  -26.71% 
 forces_ewald                          26,870,666        21,139,494            -5,731,172  -21.33% 
 forces_wolf                           663,667           642,361                  -21,306   -3.21% 
 forces_wolf                           1,017,002         994,137                  -22,865   -2.25% 
 virial                                3,173,557         3,699,969                526,412   16.59% 
 virial                                1,531,908         1,570,691                 38,783    2.53% 
 virial_ewald                          61,690,550        55,039,292            -6,651,258  -10.78% 
 virial_ewald                          90,862,664        81,927,863            -8,934,801   -9.83% 
 virial_wolf                           701,278           663,961                  -37,317   -5.32% 
```

They look very weird: on one hand we win 25% for Ewald forces, but on the other hand we loose 20% for the standard pair cache (also using an Array2 to store pair interactions). This should be worth investigating.

antoinewdg:
I think the benchmarks may just not be that reliable...

Luthaf:
Maybe, I ran them three time each.

What makes you think so?

antoinewdg:
Well in #125 I got a 25% difference for no apparent reason, at here you also have a very surprising result, it smells :(.

About the unsafe stuff, it's not using unsafe that bothers me, it's hiding its use inside the `Index` trait, and thus breaking the "segfault free" guarantees that Rust provides.

Luthaf:
> About the unsafe stuff, it's not using unsafe that bothers me, it's hiding its use inside the Index trait, and thus breaking the "segfault free" guarantees that Rust provides.

Thinking more about this, maybe we can call the wrapper `UnsafeArrayN`? And use it only where it is needed.

> Well in #125 I got a 25% difference for no apparent reason, at here you also have a very surprising result, it smells :(.

I think this is due to Travis sharing build VM on a single physical server: there was a 25% on the first set of run, but on the ones with threshold there was only a bit of differences in the cache benchmarks. I restarted the Travis run to check this hypotheses. I will also re-run my safe/unsafe benchmarks on another machine to check.

This is a downside of Travis, I don't really know how we can mitigate it. Running benchmarks there is still very desirable, maybe we can average multiple run?

Anyway, benchmarking is hard =/

Luthaf:
New result on a different processor:

```
$ cargo benchcmp --threshold=2 safe.log unsafe.log
 name                                  safe.log ns/iter  unsafe.log ns/iter  diff ns/iter   diff %
 cache_move_all_rigid_molecules        1,095,359         1,059,095                -36,264   -3.31%
 cache_move_all_rigid_molecules_ewald  8,060,593         7,577,300               -483,293   -6.00%
 cache_move_all_rigid_molecules_wolf   7,316,571         7,561,870                245,299    3.35%
 cache_move_particle                   189,242           170,931                  -18,311   -9.68%
 cache_move_particle_ewald             432,058           327,071                 -104,987  -24.30%
 cache_move_particle_wolf              93,468            91,486                    -1,982   -2.12%
 cache_move_particles                  246,347           253,027                    6,680    2.71%
 cache_move_particles_ewald            837,165           713,562                 -123,603  -14.76%
 energy_ewald                          662,784           576,538                  -86,246  -13.01%
 energy_ewald                          1,060,271         907,912                 -152,359  -14.37%
 energy_wolf                           456,610           470,129                   13,519    2.96%
 forces_ewald                          16,995,679        14,121,569            -2,874,110  -16.91%
 forces_ewald                          24,673,089        20,696,680            -3,976,409  -16.12%
 forces_wolf                           803,028           825,513                   22,485    2.80%
 virial                                1,399,952         1,442,604                 42,652    3.05%
 virial_ewald                          58,265,344        55,404,503            -2,860,841   -4.91%
 virial_ewald                          86,958,029        81,893,526            -5,064,503   -5.82%
 virial_wolf                           539,523           568,399                   28,876    5.35%
 virial_wolf                           805,988           826,382                   20,394    2.53%
```

So here I see no lost for the pair cache, but an improvement of 15% for ewald due to unchecked indexing. I think part of the issue is that benchmarks have a relatively high variance. I asked [/r/rust](https://www.reddit.com/r/rust/comments/62nyv4/tricks_and_tips_for_benchmarking_rust_code/) for help, maybe there is something obvious we did not see!

antoinewdg:
> Thinking more about this, maybe we can call the wrapper UnsafeArrayN? And use it only where it is needed.

Can't we just use `uget` of create a `at` method that is `unsafe` ? I just really don't want to hide from the compiler that we are making unsafe stuff. 

I suggest we stop the Travis discussion in this PR and keep it in #125, it's starting to get hard to keep up if it's spread between PRs.

Luthaf:
Ok, let's move the `unsafe` discussion to #130.

Issue #117, opened 2017-03-13
Broken inline math in the manual


antoinewdg:
Right now all the inline math in the manual does not work, which is not surprising since it does not use the syntax [indicated in the mdBook docs](http://azerupi.github.io/mdBook/format/mathjax.html). Did it work before, has there been breaking changes in mdBook ?

Luthaf:
This is https://github.com/azerupi/mdBook/issues/227. We configure mathjax to use `$` [here](https://github.com/lumol-org/lumol/blob/528f2d71343c2b3a1c269e543acbc785b4fce233/doc/src/theme/index.hbs#L22-L25), but out custom theme is not picked up by mdbook.

antoinewdg:
So do we actually move the `theme` back into the root `doc` folder ?

Luthaf:
Let's wait a few days to see how this is resolved on the mdbook side, and then fix on our side!

PR #122, opened 2017-03-22
Fix doc theme

antoinewdg:
#117 isn't fixed (and seemingly won't be since the `mdbook` team agreed it's better this way) so here it is.


Luthaf:
Let's go for this, and close #117. If this changes on the mdbook side, we can revert this commit.

PR #127, opened 2017-03-26
Use fixed RNG seeds in benchmarks

antoinewdg:
All is said in the title, this reduces the huge variability of cache benchmarks.


Luthaf:
I like the fact that you used separated seeds for different benchmarks here !

PR #129, opened 2017-03-28
Add a sanity check for positions in the system

Luthaf:
I had some simulations blowing up today, so I though this could be nice to have.

Remaining questions are:
1. should the checking interval be configurable by the user?
2. should we have a fully featured trait for this?

I think the answers are no and no, because (1) this should be transparent for the user (no setup required) and (2) it seems overkill (I can't think of a way someone would like to add other sanity checks from outside the core crate).

Both questions could also be addressed by (ab)using the `Output` trait for this, and maybe having 
some default output included with every simulation. They would need to be opt-out.

### DIFF
--- a/src/core/src/sim/simulations.rs
+++ b/src/core/src/sim/simulations.rs
@@ -2,6 +2,7 @@
 // Copyright (C) 2015-2016 Lumol's contributors — BSD license
 
 use sys::System;
+use types::Vector3D;
 
 use sim::Propagator;
 use sim::TemperatureStrategy;
@@ -76,11 +77,15 @@ impl Simulation {
         }
 
         self.setup(system);
-        for _ in 0..nsteps {
+        for i in 0..nsteps {
             self.propagator.propagate(system);
             system.increment_step();
             for output in &mut self.outputs {
                 output.write(system);
+            }
+
+            if i % 10000 == 0 {
+                self.sanity_check(system);
             }
         }
         self.finish(system);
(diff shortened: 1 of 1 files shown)


g-bauer:
I share your intuition that it should not be a trait. 

Would it be worthwhile to check the forces or velocities instead of the positions? A sanity check of positions would always succeed if we'd e.g. wrap molecules from time to time.

Luthaf:
We can check everything, that's right. How about 1000 A/fs as the maximal velocity and 1000 `A/fs^2/100 mass unit` for the force?

antoinewdg:
Is there a specific reason this is in `Simulation` and not in `System::increment_step()` ?

Luthaf:
I prefer to use simulation for behaviour and system for data, and the function name `increment_step` indicate this is not doing a lot of work.

antoinewdg:
My bad, I thought this was the one doing all the word, but it's `Propagator::propagate()`.

Luthaf:
I added the check on velocities too, but not on the forces as we don't store them in the system.

This is good for me.

g-bauer:
Maybe we should change the message for velocities so that we know which check failed? Maybe something as simple as "Observed very high velocities, the simulation might blow up!"

Luthaf:
Yes, we should. This is a copy/paste error on my side. That is what I get for doing too many things at the same time !

g-bauer:
Fine for me. Can I merge?

Luthaf:
Yes, sure !

To prevent ambiguity about merging, how about we stick to the following convention? We use [WIP] in the PR title if we still need/want to work on the code, and else the reviewer is free to merge whenever he wants. What do you think?

g-bauer:
Sounds good! Let's do it that way.

Issue #59, opened 2016-11-28
Add cut off distance to `resize` MC move to catch too small volumes


g-bauer:
If the box gets too small within a NPT simulation, the cut off radius will be larger than half the minimum box length (`rc > L_min/2`). If that is the case, the simulation should stop with an error.

How to get the largest cut off out of `PairInteractions`?

Luthaf:
To get the cutoff for all pair interaction, one would need to iterate over the values in [`Interactions::pairs`](https://github.com/lumol-org/lumol/blob/677d697b91ba41f0ef6c578968cf1ad8d551df21/src/core/src/sys/interactions.rs#L78), and access the cutoff from here.

The fact that there are both a `System::interaction()` and a `System::interactions_mut()` look like a code smell to me, and maybe we want to have the `System::interactions` field public, or/and move the `ParticleKinds` out of Interactions. We need to think a bit about what are the actual invariant of the `System`, `Interactions` and `ParticleKinds` structs here, to make sure nothing can go wrong after that. They are all linked by `ParticleKind`, which the user should never be able to modify directly. 

Contrary to what I said on Gitter, this does not prevent us from implementing a search for the smallest cutoff radius right now. This could be done right now by adding `all_pairs(&self) -> Vec<&[PairInteraction]>` to `Interactions`, and `PairInteraction::cutoff(&self) -> f64` functions. Then we could look for the smallest cutoff in `McMove::setup` function.

g-bauer:
> Then we could look for the smallest cutoff in McMove::setup function.

I had this wrong. We need to check for the *largest* cut off. I thought of asking for the cut off when calling `Resize::new()` but maybe it is better in `Resize::prepare` since we have access to the `system`.

Luthaf:
>  Doing so instead of McMove::setup limits this case to Resize which eventually makes the McMove routines more general and readable.

Sorry, I though there was a `McMove::setup` function, like the [`Integrator` trait](https://github.com/lumol-org/lumol/blob/677d697b91ba41f0ef6c578968cf1ad8d551df21/src/core/src/sim/md/integrators.rs#L10-L12). My issue with `prepare` is that is is called at every step, while `setup` would be called only once by simulation run.

We should add this function to the `McMove` trait, with a default empty implementation, and call it from `MonteCarlo::setup` for all the moves in the simulation. This would still be limited to Resize, and called only once by simulation.

g-bauer:
> My issue with prepare is that is is called at every step

[...]

PR #134, opened 2017-04-03
Add function to get cutoff radius for global potentials.

g-bauer:
This PR adds a function to `GlobalPotentials`, that returns their `cutoff`.
In contrast to `PairPotentials`, `GlobalPotentials` may not have a cutoff so that the function returns an `Option`.

### DIFF
--- a/src/core/src/energy/global/ewald.rs
+++ b/src/core/src/energy/global/ewald.rs
@@ -704,6 +704,10 @@ impl Ewald {
 }
 
 impl GlobalPotential for Ewald {
+    fn get_cutoff(&self) -> Option<f64> {
+        Some(self.rc)
+    }
+
     fn energy(&mut self, system: &System) -> f64 {
         self.precompute(system.cell());
         let real = self.real_space_energy(system);
--- a/src/core/src/energy/global/ewald.rs
+++ b/src/core/src/energy/global/ewald.rs
@@ -704,7 +704,7 @@ impl Ewald {
 }
 
 impl GlobalPotential for Ewald {
-    fn get_cutoff(&self) -> Option<f64> {
+    fn cutoff(&self) -> Option<f64> {
         Some(self.rc)
     }
 
--- a/src/core/src/energy/global/mod.rs
+++ b/src/core/src/energy/global/mod.rs
@@ -39,6 +39,10 @@ use energy::PairRestriction;
 /// }
 ///
 /// impl GlobalPotential for ShiftAll {
+///     fn get_cutoff(&self) -> Option<f64> {
+///         None
+///     }
+///
 ///     fn energy(&mut self, system: &System) -> f64 {
 ///         // shift all particles by delta
 ///         self.delta * system.size() as f64
@@ -84,6 +88,8 @@ use energy::PairRestriction;
 /// assert_eq!(system.virial(), Matrix3::zero());
 /// ```
 pub trait GlobalPotential: GlobalCache + BoxCloneGlobal {
+    /// Return the cut off radius.
+    fn get_cutoff(&self) -> Option<f64>;
     /// Compute the energetic contribution of this potential
     fn energy(&mut self, system: &System) -> f64;
     /// Compute the force contribution of this potential. This function should
@@ -120,6 +126,10 @@ impl_box_clone!(GlobalPotential, BoxCloneGlobal, box_clone_gobal);
 /// }
 ///
 /// impl GlobalPotential for ShiftAll {
+///     fn get_cutoff(&self) -> Option<f64> {
+///         None
+///     }
+///
 ///     fn energy(&mut self, system: &System) -> f64 {
 ///         // shift all particles by delta
 ///         self.delta * system.size() as f64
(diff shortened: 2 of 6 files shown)


src/core/src/energy/global/ewald.rs
```diff
@@ -704,6 +704,10 @@ impl Ewald {
 }
 
 impl GlobalPotential for Ewald {
+    fn get_cutoff(&self) -> Option<f64> {
```
Luthaf: Can we name this function `cutoff`? The usual way in Rust is to have getter called `foo` and setters called `set_foo`. Here is the corresponding [RFC](https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis).

Luthaf:
Do you want to tackle #59 here too, or should it be part of a separated PR?

src/core/src/energy/global/ewald.rs
g-bauer: Sure. I am using the same name here that we have for the pair potential. Might change it as well.

g-bauer:
> Do you want to tackle #59 here too, or should it be part of a separated PR?

I already have the follow up PR in line, if that's ok.

Luthaf:
> I already have the follow up PR in line, if that's ok.

No problem!

antoinewdg:
I'm confused, I thought we were rebasing to avoid merge commits (both last PRs created a merge commit) ?

g-bauer:
Can you clarify what you mean by that for a git-noob? 😄

Luthaf:
> I'm confused, I thought we were rebasing to avoid merge commits (both last PRs created a merge commit) ?

I am not sure rebasing prevent merge conflicts. It prevent merge commits, but the changes to the files are the same, aren't they?

> Can you clarify what you mean by that for a git-noob? 😄

Rebasing means taking the branch and applying it commit by commit on top of another branch, here master. Github offer both a 'Merge PR' option and a 'Rebase PR' option. Merge conflicts happen when two commits changes the same files at the same place, and git does not know how which changes to use, so it bails out and ask the user to resolve the conflicts.

You can usually resolve merge conflicts by adding a new commit, or by rebasing your branch on top of the conflicting branch, and modifying/adapting your commits on the go. See https://git-scm.com/book/en/v2/Git-Branching-Rebasing for more informations.

-----

I usually use 'rebase and merge' for smaller PR (1-2 commits), and 'Merge' otherwise; but we can settle on something else. I just don't like the 'Squash and Merge' option, I prefer to ask the OP to squash if this is needed.

antoinewdg:
> I am not sure rebasing prevent merge conflicts. It prevent merge commits, but the changes to the files are the same, aren't they?

Yes exactly (I did say merge commits :P). 
The advantage of rebase is that it keeps a flat commit history and:
- you said you wanted a clean history
- I can see the benchmarks regularly run on master in the future, a flat history tree would help


As an illustration of the difference between rebase and merge:
```bash
$ git log --graph --oneline # on master
*   0649f740 Merge pull request #134 from g-bauer/global_potential_cutoff
|\  
| * 24014d06 Changed function name.
| * 30b27a2b Add function to get cutoff radius for global potentials.
* |   9337ddfc Merge pull request #129 from lumol-org/sanity
|\ \  
| |/  
|/|   
| * 12cb4d4b Add a sanity check for positions and velocities in the system
* | 805fe938 Update TOML to 0.3
* | 0ad212a5 Update lumol-core dependencies
* | 740eee07 Use fixed RNG seeds in benchmarks
|/  
* fd0a9efa Uniformize memory access
* 6da0482f Move the computation of fourier_i one up in the loop
* d241f590 Remove useless computation of real part of phase in Ewald::kspace_force_factor
* 83ba3acc Remove redundant multiplications
* f6e042dc Refactor similar loops between virial and force in kspace for Ewald
* 86b59d20 Fix doc theme
* 431fca20 Reorganise tests data files
* 1adba6a8 Check averaged values in NPT simulations
* 5b523acb We only compute energy for the big nacl crystal with Ewald
```

The bottom part is what happens when we rebase, the top when we merge.

@g-bauer if you feel like learning more about git, I learned it [there](https://www.atlassian.com/git/tutorials/what-is-version-control) and really like how they teached it, [here](https://www.atlassian.com/git/tutorials/merging-vs-rebasing) is their explanation of merge/rebase. The only thing I can reproach is that they hid the rebase stuff in "Advanced tips" and I wish a knew about it way before :).

g-bauer:
> Rebasing means taking the branch and applying it commit by commit on top of another branch, here master.

Ok but would that change anything for this case here? I usually fetch the current master of this repo and locally rebase my branch on top. So what you are discussing is only a problem when master changed after PR, right?

g-bauer:
@antoinewdg Thanks! I missed that I merged the other PR after I opened this PR (for which I rebased beforehand). Oooops.

antoinewdg:
If you do the rebasing locally it shouldn't matter, but if you use the rebase button on Github it's a failsafe for when you forget. 

> So  what you are discussing is only a problem when master changed after PR, right?

Right, but this happens a lot.

g-bauer:
Ok thanks, I'll keep that in mind for the next PR.

Luthaf:
> Yes exactly (I did say merge commits :P).

Sorry, I misread your comment. 

Having a clean history is nice to have, we can stick to rebase and merge for now. But it might be easier to switch to simple merge if we ever get a lot of PR at the same time someday =)

PR #140, opened 2017-04-04
Npt MC check global cutoff in setup function.

g-bauer:
Adds checking of cutoffs (If present) from `GlobalPotential` within `setup` in the `Resize` MC move.

### DIFF
--- a/src/core/src/sim/mc/moves/resize.rs
+++ b/src/core/src/sim/mc/moves/resize.rs
@@ -22,8 +22,8 @@ pub struct Resize {
     new_system: System,
     /// target pressure
     pressure: f64,
-    /// largest cut off diameter of `PairPotentials`
-    rc_max: f64,
+    /// largest cutoff diameter of potentials in `Interactions`
+    maximum_cutoff: Option<f64>,
 }
 
 impl Resize {
@@ -36,7 +36,7 @@ impl Resize {
             range: Range::new(-delta, delta),
             new_system: System::new(),
             pressure: pressure,
-            rc_max: 0.0,
+            maximum_cutoff: None,
         }
     }
 }
@@ -54,23 +54,7 @@ impl MCMove for Resize {
 
         // Get the largest cutoff of all intermolecular interactions in the
         // system.
-
-        // Go through global interactions
-        let rc_glob = system.interactions()
-                            .globals()
-                            .iter()
-                            .map(|i| i.cutoff())
-                            .filter_map(|rc| rc)
-                            .fold(f64::NAN, f64::max);
-
-        // Pair interactions
-        let rc_pairs = system.interactions()
-                            .all_pairs()
-                            .iter()
-                            .map(|i| i.cutoff())
-                            .fold(f64::NAN, f64::max);
-
-        self.rc_max = f64::max(rc_glob, rc_pairs)
+        self.maximum_cutoff = system.interactions().maximum_cutoff()
     }
 
     fn prepare(&mut self, system: &mut System, rng: &mut Box<Rng>) -> bool {
@@ -88,14 +72,17 @@ impl MCMove for Resize {
         // Check the radius of the smallest inscribed sphere and compare to the
         // cut off distance.
         // Abort simulation when box gets smaller than twice the cutoff radius.
-        if self.new_system.cell()
-            .lengths()
-            .iter()
-            .any(|&d| 0.5 * d <= self.rc_max) {
-            fatal_error!("Tried to decrease the cell size but new size
-                conflicts with the cut off radius. \
-                Increase the number of particles to get rid of this problem.")
-        }
+        if let Some(maximum_cutoff) = self.maximum_cutoff {
+            if self.new_system.cell()
+                    .lengths()
+                    .iter()
+                    .any(|&d| 0.5 * d <= maximum_cutoff) {
+                    fatal_error!("Tried to decrease the cell size but new size
+                        conflicts with the cut off radius. \
+                        Increase the number of particles to get rid of this problem.")
+                    }
+        };
+
         // Loop over all molecules in the system.
         // We don't want to change the intramolecular distances
         // so we compute the translation vector of the center-of-mass
--- a/src/core/src/sim/mc/moves/resize.rs
+++ b/src/core/src/sim/mc/moves/resize.rs
@@ -54,12 +54,23 @@ impl MCMove for Resize {
 
         // Get the largest cutoff of all intermolecular interactions in the
         // system.
-        // TODO: include electrostatic interactions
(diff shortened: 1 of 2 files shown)


src/core/src/sim/mc/moves/resize.rs
```diff
@@ -54,12 +54,23 @@ impl MCMove for Resize {
...
+
+        // Pair interactions
+        let rc_pairs = system.interactions()
                             .all_pairs()
                             .iter()
                             .map(|i| i.cutoff())
-                            .fold(f64::NAN, f64::max)
+                            .fold(f64::NAN, f64::max);
+
+        self.rc_max = f64::max(rc_glob, rc_pairs)
```
Luthaf: Nit-picking on names: `self.max_cutoff = f64::max(global_cutoff, pairs_cutoff)`

src/core/src/sim/mc/moves/resize.rs
```diff
@@ -54,12 +54,23 @@ impl MCMove for Resize {
...
-        // TODO: include electrostatic interactions
-        self.rc_max = system.interactions()
+
+        // Go through global interactions
+        let rc_glob = system.interactions()
+                            .globals()
+                            .iter()
+                            .map(|i| i.cutoff())
+                            .filter_map(|rc| rc)
+                            .fold(f64::NAN, f64::max);
```
Luthaf: What happens if there are no cutoff anywhere? Then we get `self.rc_max = f64::NAN)`, and then ?

src/core/src/sys/interactions.rs
```diff
@@ -443,4 +445,17 @@ mod test {
         interactions.add_global(Box::new(Wolf::new(1.0)));
         assert_eq!(interactions.globals().len(), 1);
     }
+
+    #[test]
+    fn test_globals_cutoff() {
```
Luthaf: Why is this test needed? We should have tests on `Resize` instead, checking that we get the right cutoff in specific cases:

- no cutoff
- only pairs
- only globals
- both

Or we could make this a function of `Interactions` and test it here.

antoinewdg: +1 for making it part of `Interaction` because this is very likely to be reused later.

src/core/src/sim/mc/moves/resize.rs
g-bauer: Should we do this globally then, call it `cutoff` instead of `rc`? We have a variety of different names currently.

src/core/src/sys/interactions.rs
g-bauer: If we decide to make it a function of `Interactions`, what functionality would be good? Return the largest cutoff? Return a list of all cutoffs?

src/core/src/sim/mc/moves/resize.rs
Luthaf: Sure, we can use `cutoff` everywhere.

src/core/src/sys/interactions.rs
Luthaf: I think we mainly need to get the largest cutoff. Let's use the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle!

src/core/src/sim/mc/moves/resize.rs
Luthaf: Yes, that's what I meant. Do we handle the case where there is no cutoff when comparing in the Resize move?

g-bauer: The check does the following:
```rust
if self.new_system.cell()                                                                                                                                       
            .lengths()
            .iter()
            .any(|&d| 0.5 * d <= self.rc_max) {
                fatal_error!("Tried to decrease the cell size but new size
                conflicts with the cut off radius. \
                Increase the number of particles to get rid of this problem.")
           }
```
If I understand the behavior of `NAN` correctly, every f64 (negative as well as positive) should be larger than `NAN`. So even if `self.rc_max = f64::NAN`, the error should not be triggered. But that is not very transparent. We could skip the whole if case by checking `if self.rc_max.is_nan`.

Am I getting this right? What do you think?

src/core/src/sys/interactions.rs
g-bauer: OK! Then I'll add a function that returns the largest cutoff. I think returning an `Option<f64>` would be the most transparent way to do it. Should I add it here or open a new PR?

src/core/src/sim/mc/moves/resize.rs
g-bauer: If we decide to write a function `maximum_cutoff` in `Interactions`, I'd return an `Option<f64>`. In that case, we'd have to catch if there is a cutoff explicitly.

src/core/src/sys/interactions.rs
Luthaf: Add it to this PR, no need to add complexity.

src/core/src/sys/interactions.rs
```diff
@@ -274,15 +275,37 @@ impl Interactions {
     pub fn globals(&self) -> &[Box<GlobalPotential>] {
         &self.globals
     }
+
+    /// Get maximum cutoff from all `pairs` and all `globals`.
+    pub fn maximum_cutoff(&self) -> Option<f64> {
+        // Go through global interactions
```
Luthaf: This also need to check for the coulombic cutoff (here and in tests)

g-bauer:
Holy cow, I feel like the function is poorly written.

Luthaf:
> Holy cow, I feel like the function is poorly written.

Why?

g-bauer:
> Why?

The whole `None => f64::NAN` feels weird. But hey, it works.

Luthaf:
OK, this is good for me. 

Could you [squash related commits](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history) together?

g-bauer:
Done! Hope it's ok this way.

Luthaf:
Great, thanks!

Luthaf:
cr*p I messed up the button again and used a plain merge instead of rebase and merge =/

I should go to sleep ...

g-bauer:
Can we undo that? Is it needed? I could restore ...

Luthaf:
We could undo, but I prefer starting to treat master as immutable =)

It is not a major issue, just something a bit annoying. I should be more careful, but don't worry about this!

g-bauer:
I am still confused ... I rebased before pushing. Would it have made a difference in this case?

Luthaf:
No, because github use `git merge --no-ff`. Which means it always create a merge commit, even if the branch is up to date with the latest master.

PR #141, opened 2017-04-05
Logging configuration in the input files

### LINKED ISSUE #110: Logging configuration in the input file
The logging should be configurable in the input file. The user could specify whether the logging should be send to a file, stdout, stderr or a combination of these. 

We could use the [simplelog](https://docs.rs/simplelog/) crate to provide the logger implementation, and remove the one in `src/core/src/logging.rs`.

I think that default should be to log errors/warnings to stderr and everything else to stdout.

Any idea on how the input file could look like for this?

Luthaf:
Fix #110. ~~This still needs doc in the user manual.~~

Doing this because I didn't want to add API documentation to the previous in-tree Logger.

I used [log4rs](https://crates.io/crates/log4rs) instead of simplelog, because it allow more configuration to the message formatting. log4rs is a bit harder to use (feels like Java), but more powerful.

This is the output when writing to a console:
logging

Logging to a file is the same, without any color.

## Syntax

I went with the following syntax:

```toml
# Multiple targets
[log]
targets = [
    {target = "<stdout>", level = "warning"},
    {target = "file.log", append = true},
]

# Single target (1)
[log]
target = "<stderr>"

# Single target (2)
[log]
target = "lumol.log"
level = "trace"
append = false
```

## Default logger

If nothing is specified in the input file, a default logger, equivalent to `{target = "<stdout>", level = "info"}` is initialized. If a logger was already initialized (because the user wrote his own), the input file logger is (silently) ignored.


g-bauer:
That's nice! What do you think about having both stdout and writing to a file (e.g. lumol.log) as default if nothing is specified?

Luthaf:
Why not, it can be nice. We don't output all the simulation data to the log file (contrary to a lot of software), so keeping the log file around is not as necessary as it can be for other software. 

But we are currently only writing to the disk what was requested by the user. Adding another implicit file may make it a bit harder to use lumol as a library, inside other software.

antoinewdg:
Nice ! 

> But we are currently only writing to the disk what was requested by the user. Adding another implicit file may make it a bit harder to use lumol as a library, inside other software.

I agree on this point.

Luthaf:
I added the doc, so this is complete for me.

doc/src/input/log.md
```diff
@@ -0,0 +1,81 @@
+# Logging configuration
+
+Lumol send various logging messages while running a simulation. Some of them are
```
antoinewdg: sends

doc/src/input/log.md
```diff
@@ -0,0 +1,81 @@
...
+informational messages (`charge was set to 1.2 for 132 particles`), others are
+warnings or error message (`infinite energy!`) and some are for debug purposes.
+
+By default, the code print all the informational, warning and error messages to
```
antoinewdg: prints
Maybe use Lumol again instead of "the code" ?

doc/src/input/log.md
```diff
@@ -0,0 +1,81 @@
...
+the standard terminal output. This allow to run the code and redirect the output
```
antoinewdg: allows

doc/src/input/log.md
```diff
@@ -0,0 +1,81 @@
...
+to a specific file with the usual UNIX way: `lumol input.toml > simulation.log`.
+
+Lumol also offer finer-grain configuration for logging output, for example only
```
antoinewdg: offers

doc/src/input/log.md
```diff
@@ -0,0 +1,81 @@
...
+printing errors and warnings, and redirecting everything else to a file. This
+configuration happens in the `[log]` section of the input file. This section can
+contains either a single output `target`, or multiple `targets`.
```
antoinewdg: contain


image

doc/src/input/log.md
```diff
@@ -0,0 +1,81 @@
...
+Optional keys are `level` and `append`. The `level` key controls which messages
+are sent to this output, and default to `info`. Available levels are the
+following:
+
+- `error`: only error messages;
+- `warning`: error and warning messages;
+- `info`: error, warnings and informational messages;
+
+The `debug` (debug messages) and `trace` (very verbose debug) are also
+available, but mainly intended for development of the code.
```
antoinewdg: s/development of the code/developers ?

doc/src/input/log.md
```diff
@@ -0,0 +1,81 @@
...
+
+The `append` key is a Boolean value (`true` or `false`) only used for files, and
+controlling whether to overwrite the file or append new messages at the end of
+the file. It default to `false`, meaning that the file is overwritten by every
```
antoinewdg: defaults

doc/src/input/log.md
Luthaf: I am really bad at putting the 's' at the third person =/ Thank you for the review!

PR #139, opened 2017-04-04
[WIP] Script for running benchmarks automatically

antoinewdg:
This PR presents a script used to run the benchmarks automatically in for a PR.

```
$ python scripts/run-benchmarks.py --help
Usage: run-benchmarks.py [OPTIONS] OUTPUT_DIR

  Run the benchmarks.

  If --pr-id is set, runs the benchmarks from this PR. The number of commits
  to bench is either the number provided in the PR body using [lumol bench
  <n_commits>], or as a parameter to this script (in case of both the script
  parameter prevails). The benchmark results are saved in OUTPUT_DIR, and
  if a PR id have been set a comment is automatically added to it with the
  results of the PR.

Options:
  -n, --n-commits INTEGER  number of commits to benchmark
  -p, --pr-id INTEGER
  --help                   Show this message and exit.
```

The script is automatically run on PR on Travis CI, if the number of commits is set in the PR body using [lumol bench <n>].

[lumol bench 2]

PR #143, opened 2017-04-08
Run benchmarks locally for a PR

antoinewdg:
This PR introduces a script that runs the benchmarks for a PR.

```
$ python scripts/run-benchmarks.py --help
Usage: run-benchmarks.py [OPTIONS] OUTPUT_DIR N_COMMITS PR_ID

  Run the benchmarks for multiple commits on a PR and compare to master.

  The benchmark results are saved in OUTPUT_DIR, and a comment with a
  summary will be automatically added to the PR. This script requires the
  environment variables LUMOL_GH_USERNAME and LUMOL_GH_TOKEN to contain
  respectively the Github username and a personal access token.

Options:
  --help  Show this message and exit.

```

After spending far too much time trying to design a script for this, I went with minimal functionality of running the benchmarks for a specific PR, for a specific number of commits.
(#139 comfirmed that benchmarks on Travis will not happen)

### DIFF
--- a/scripts/run-benchmarks.py
+++ b/scripts/run-benchmarks.py
@@ -1,0 +1,185 @@
+import subprocess
+import json
+import os
+import re
+
+import click
+import cpuinfo
+
+BASE_REPO_URL = 'https://github.com/lumol-org/lumol.git'
+
+INDIVIDUAL_BENCH_TEMPLATE = """
+<details><summary>%s %s</summary>
+  <p>\n\n
+\n\n
+```bash\n
+%s\n\n
+```\n\n
+
+</p></details>
+"""
+
+COMPARISON_TEMPLATE = """
+<details><summary>%s %s</summary>
+  <p>\n\n
+\n\n
+```bash\n
+%s\n\n
+```\n\n
+
+</p></details>
+"""
+
+
+def request_api(endpoint, data=None):
+    url = 'https://api.github.com/repos/lumol-org/lumol' + endpoint
+    username = os.environ['LUMOL_GH_USERNAME']
+    token = os.environ['LUMOL_GH_TOKEN']
+
+    cmd = "curl -sS %s  -u '%s:%s'" % (url, username, token)
+    if data is not None:
+        cmd += " --data '%s'" % json.dumps(data)
+
+    out = subprocess.check_output(cmd, shell=True).decode('utf-8')
+    return json.loads(out)
+
+
+def get_master_commit():
+    cmd = 'git log --oneline upstream/master -n 1'
+    out = subprocess.check_output(cmd, shell=True).decode('utf-8')
+    h, _, title = out.rstrip().partition(' ')
+    return h, title
+
+
+def get_commit_descriptions(n_commits):
+    """
+    Get hash and title of the `n_commits` latest commits on the branch.
+
+    Also adds the commit at the HEAD of master in the end. If this
+    commit is met earlier, stops at this commit (the master commit
+    is guaranteed to be at the end of the result.
+
+    :param n_commits:
+    :return:
+    """
+    cmd = 'git log --oneline | head -n %s' % n_commits
+    out = subprocess.check_output(cmd, shell=True).decode('utf-8')
+    master_h, master_title = get_master_commit()
+
+    descriptions = []
+    for line in out.split('\n'):
+        if line.rstrip() == '':
+            continue
+        h, _, title = line.partition(' ')
+        descriptions.append((h, title))
+        if h == master_h:
+            break
+
+    if descriptions[-1][0] != master_h:
+        descriptions.append((master_h, master_title))
+
+    return descriptions
+
+
+def setup_cloned_repo(pr_id):
+    response = request_api('/pulls/%s' % pr_id)
+    clone_url = response['head']['repo']['clone_url']
+    commit_id = response['head']['sha']
+
+    subprocess.call('git clone %s cloned_repo' % clone_url, shell=True)
+    os.chdir('cloned_repo')
+    subprocess.call('git checkout %s' % commit_id, shell=True)
+    subprocess.call('git remote add upstream %s' % BASE_REPO_URL, shell=True)
+    subprocess.call('git fetch upstream master', shell=True)
+
+
+class Benchmarker:
+    def __init__(self, n_commits, output_dir):
+        self.commit_descriptions = get_commit_descriptions(n_commits)
+        self.output_dir = os.path.abspath(os.path.expanduser(output_dir))
+
+    def run_warmup(self):
+        print('=================== Warming up ==============================')
+        for _ in range(3):
+            subprocess.check_output('cargo bench', shell=True)
+
+    def run_bench(self, h):
+        cmd = 'cargo bench > %s/%s.txt' % (self.output_dir, h)
+        subprocess.call(cmd, shell=True)
+
+    def run_all_benches(self):
+        for h, title in self.commit_descriptions:
+            print('=================== Benching commit %s ==============================' % h)
+            subprocess.call('git checkout %s' % h, shell=True)
+            self.run_bench(h)
+            print('=================== Done ==============================')
+
+    def compare_benches(self):
+        comparisons = {}
+        master_h, _ = self.commit_descriptions[-1]
+        for h, title in self.commit_descriptions[:-1]:
+            cmd = 'cargo benchcmp %s/%s.txt %s/%s.txt --threshold 2 --variance' % (
+                self.output_dir, master_h, self.output_dir, h
+            )
+            out = subprocess.check_output(cmd, shell=True).decode('utf-8')
+            comparisons[h] = out
+
+        return comparisons
+
+    def comment_pr(self, pr_id):
+        # Comparison benchmarks
+        master_h, master_title = self.commit_descriptions[-1]
+        comment = '## Comparing to master (%s)\nusing `--threshold 2, latest commit first`' % master_h
+
+        comparisons = self.compare_benches()
+        for h, title in self.commit_descriptions[:-1]:
+            compare = comparisons[h]
+            comment += COMPARISON_TEMPLATE % (h, title, compare)
+
+        # Individual benchmarks
+        comment += '\n## Individual benchmarks\n'
+
+        for k, (h, title) in enumerate(self.commit_descriptions):
+            with open('%s/%s.txt' % (self.output_dir, h)) as f:
+                bench = f.read()
+            comment += INDIVIDUAL_BENCH_TEMPLATE % (h, title, bench)
+
+        info = cpuinfo.get_cpu_info()
+        if info is not None:
+            comment += '\n<br>**CPU**: %s' % info['brand']
+
+        # Emit the request
+        data = {
+            'body': comment
+        }
+        request_api('/issues/%s/comments' % pr_id, data)
+
+
+@click.command()
+@click.argument('output_dir')
+@click.argument('n_commits', type=click.INT)
+@click.argument('pr_id', type=click.INT)
+def main(output_dir, n_commits, pr_id):
+    """
+    Run the benchmarks for multiple commits on a PR and compare to master.
+
+    The benchmark results are saved in OUTPUT_DIR, and a comment
+    with a summary will be automatically added to the PR.
+    This script requires the environment variables LUMOL_GH_USERNAME
+    and LUMOL_GH_TOKEN to contain respectively the Github username
+    and a personal access token.
+    """
+
+    setup_cloned_repo(pr_id)
+
+    benchmarker = Benchmarker(n_commits, output_dir)
+    benchmarker.run_warmup()
+    benchmarker.run_all_benches()
+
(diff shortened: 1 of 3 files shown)


scripts/run-benchmarks.py
```diff
@@ -0,0 +1,185 @@
...
+```\n\n
+
+</p></details>
+"""
+
+
+def request_api(endpoint, data=None):
+    url = 'https://api.github.com/repos/lumol-org/lumol' + endpoint
+    username = os.environ['LUMOL_GH_USERNAME']
+    token = os.environ['LUMOL_GH_TOKEN']
```
Luthaf: Can we have a nice error message if the variable is not found?

scripts/run-benchmarks.py
```diff
@@ -0,0 +1,185 @@
...
+\n\n
+```bash\n
+%s\n\n
+```\n\n
+
+</p></details>
+"""
+
+COMPARISON_TEMPLATE = """
+<details><summary>%s %s</summary>
```
Luthaf: Could you use 'new' string formatting, using `"{}".format()`? It was added in [Python 2.6](https://docs.python.org/2/library/string.html#custom-string-formatting), so it is always available.

scripts/run-benchmarks.py
```diff
@@ -0,0 +1,185 @@
...
+    Get hash and title of the `n_commits` latest commits on the branch.
+
+    Also adds the commit at the HEAD of master in the end. If this
+    commit is met earlier, stops at this commit (the master commit
+    is guaranteed to be at the end of the result.
+
+    :param n_commits:
+    :return:
+    """
+    cmd = 'git log --oneline | head -n %s' % n_commits
```
Luthaf: Could we not use `head` here? It makes the code non-portable to Windows, and we can achieve the same effect with `git log --oneline -n <n>`

scripts/run-benchmarks.py
```diff
@@ -0,0 +1,185 @@
...
+    out = subprocess.check_output(cmd, shell=True).decode('utf-8')
+    master_h, master_title = get_master_commit()
+
+    descriptions = []
+    for line in out.split('\n'):
+        if line.rstrip() == '':
+            continue
+        h, _, title = line.partition(' ')
+        descriptions.append((h, title))
+        if h == master_h:
```
Luthaf: What is `h`? HEAD?

Or is it the commit sha1? In that case, I would use `ref`/`rev` instead

scripts/run-benchmarks.py
```diff
@@ -0,0 +1,185 @@
...
+
+    return descriptions
+
+
+def setup_cloned_repo(pr_id):
+    response = request_api('/pulls/%s' % pr_id)
+    clone_url = response['head']['repo']['clone_url']
+    commit_id = response['head']['sha']
+
+    subprocess.call('git clone %s cloned_repo' % clone_url, shell=True)
```
Luthaf: I am a bit uneasy on cloning the whole repository, just to run some benchmarks. What prevent running them on the local clone?

scripts/run-benchmarks.py
```diff
@@ -0,0 +1,185 @@
...
+        for _ in range(3):
+            subprocess.check_output('cargo bench', shell=True)
+
+    def run_bench(self, h):
+        cmd = 'cargo bench > %s/%s.txt' % (self.output_dir, h)
+        subprocess.call(cmd, shell=True)
+
+    def run_all_benches(self):
+        for h, title in self.commit_descriptions:
+            print('=================== Benching commit %s ==============================' % h)
```
Luthaf: s/benching/'benchmarking' or 'running benchmarks for'

scripts/run-benchmarks.py
```diff
@@ -0,0 +1,185 @@
...
+            subprocess.call('git checkout %s' % h, shell=True)
+            self.run_bench(h)
+            print('=================== Done ==============================')
+
+    def compare_benches(self):
+        comparisons = {}
+        master_h, _ = self.commit_descriptions[-1]
+        for h, title in self.commit_descriptions[:-1]:
+            cmd = 'cargo benchcmp %s/%s.txt %s/%s.txt --threshold 2 --variance' % (
```
Luthaf: Could we get a nicer error if cargo benchcmp is not installed?

scripts/run-benchmarks.py
```diff
@@ -0,0 +1,185 @@
...
+</p></details>
+"""
+
+
+def request_api(endpoint, data=None):
+    url = 'https://api.github.com/repos/lumol-org/lumol' + endpoint
+    username = os.environ['LUMOL_GH_USERNAME']
+    token = os.environ['LUMOL_GH_TOKEN']
+
+    cmd = "curl -sS %s  -u '%s:%s'" % (url, username, token)
```
Luthaf: Can we have a nice error message if curl is not found? 

Or if you have the motivation (I would understand if not ^^), could we use the `request` Python package here?

Luthaf:
>  (#139 comfirmed that benchmarks on Travis will not happen)

Too much variability ?

antoinewdg:
> Too much variability ?

Yes if you look at the  [latest comment](https://github.com/lumol-org/lumol/pull/139#issuecomment-291552974) of the lumol bot there is some 60% variability that has no explanation.

scripts/run-benchmarks.py
antoinewdg: If you try to benchmark a PR initiated by me, you won't find my branch in your repo (because it comes from a fork). Are there cases where the bandwidth used to clone repetitively can be an issue ?

scripts/run-benchmarks.py
antoinewdg: I used `h` for "hash" (so yes the commit sha1), just realized it does not work in English ! I'll change that.

scripts/run-benchmarks.py
Luthaf: It is not an issue for now, but it might become. Did you considered using [`git fetch origin pull/<id>/head:pr-<id>`](http://stackoverflow.com/a/30584951) for this?

antoinewdg: I had no idea this existed, I'll look into it thanks !

scripts/run-benchmarks.py
antoinewdg: Yes sure, it's indeed better for long string. I still find the older nicer for smaller strings (just because `%` is shorter that `.format()`), is it a problem if we use them both ?

Luthaf: I prefer if we only use one version throughout the code, to make it consistent. The `format` and bracketed version is the 'modern' version, and main version in Python 3, that's why I am pushing for it 😃

antoinewdg:
## Comparing to master (3e54b97f)
using `--threshold 2, latest commit first`
<details><summary>99618dcd Add PR benchmarking script</summary>
  <p>





```bash




```



</p></details>

## Individual benchmarks

<details><summary>99618dcd Add PR benchmarking script</summary>
  <p>





```bash


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured


running 1 test
test energy_wolf               ... bench:     471,850 ns/iter (+/- 7,935)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured


running 1 test
test energy_wolf                          ... bench:     731,219 ns/iter (+/- 7,804)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured




```



</p></details>

<details><summary>3e54b97f Merge pull request #140 from g-bauer/npt_mc_check_global_cutoff</summary>
  <p>





```bash


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured


running 1 test
test energy_wolf               ... bench:     472,395 ns/iter (+/- 3,892)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured


running 1 test
test energy_wolf                          ... bench:     729,832 ns/iter (+/- 1,322)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured




```



</p></details>

<br>**CPU**: Intel(R) Core(TM) i7-5700HQ CPU @ 2.70GHz

antoinewdg:
@Luthaf I think I have addressed all your comments :)

scripts/run-benchmarks.py
```diff
@@ -0,0 +1,200 @@
...
+    if data is None:
+        r = requests.get(url, auth=(username, token))
+    else:
+        r = requests.post(url, auth=(username, token), json=data)
+
+    return r.json()
+
+
+def get_master_commit():
+    cmd = 'git log --oneline upstream/master -n 1'
```
Luthaf: Should this be `_bot_remote/master` instead?

scripts/run-benchmarks.py
```diff
@@ -0,0 +1,200 @@
...
+    if var is None:
+        msg = 'The environment variable {} must be set to access Github API'.format(name)
+        raise Exception(msg)
+
+    return var
+
+
+def check_for_cargo_benchcmp():
+    try:
+        subprocess.check_output('cargo benchcmp --help', shell=True)
```
Luthaf: BTW why is `shell=True` needed here?

scripts/run-benchmarks.py
```diff
@@ -0,0 +1,200 @@
...
+    return descriptions
+
+
+def clean_repo():
+    subprocess.call('git checkout master', shell=True)
+    subprocess.call('git branch -D _bot_pr', shell=True)
+    subprocess.call('git remote remove _bot_remote', shell=True)
+
+
+def setup_repo(pr_id):
```
Luthaf: Shouldn't this be a context manager? Else is can leave the repo in a dirty state if there is an exception somewhere.

As you are already cleaning before anything, this should be fine, but it might be cleaner to ensure the cleaning at the end of the script.

scripts/run-benchmarks.py
antoinewdg: Because if it's not we have to call the commands with a list. For example here `['cargo', 'benchcmp', '--help']` and getting the output is also more painful. This is best to not use `shell=True` for security reasons, I don't see it being a problem here.

scripts/run-benchmarks.py
antoinewdg: Good catch thanks

scripts/run-benchmarks.py
Luthaf: Ok, didn't knew this!

antoinewdg:
@Luthaf this should be it. If you could test it that would be awesome, because there may be some stuff that differ on your repo that I did not account for.

Luthaf:
I got an error \o/. When passing a non-existent directory as output dir, the code fails on file creation.

After testing it locally, I have a few more comments: could you make the script executable with a shebang so that we can just run `./scripts/run-benchmarks.py ...`?

Also the arguments feels strange: why do we need to specify the number of commits? I think it should default to the whole PR history. The output dir should also default to something like `target/benches`.

I would also reorganise the arguments in the [input] [output] order: `./scripts/run-benchmarks.py 131 [-n 6] [target/benches]`. The `-n` is required to differentiate between the output and the number of commits to benchmark.

Finally, it could be nice to have a "Comment posted on PR <id>" at the end of the run.

antoinewdg:
> I got an error \o/. When passing a non-existent directory as output dir, the code fails on file creation.

It's not a bug it's a feature :stuck_out_tongue: . I mean most scripts that take an output dir fail if it doesn't exist right ? Anyway you are right it would be more convenient if it did not.

> After testing it locally, I have a few more comments: could you make the script executable with a shebang so that we can just run ./scripts/run-benchmarks.py ...?

Sure

> Also the arguments feels strange: why do we need to specify the number of commits? I think it should default to the whole PR history. 

I agree.

> The output dir should also default to something like target/benches

I did not think about that, `target` is indeed the perfect target directory.

> I would also reorganise the arguments in the [input] [output] order: ./scripts/run-benchmarks.py 131 [-n 6] [target/benches]. The -n is required to differentiate between the output and the number of commits to benchmark.
>
> Finally, it could be nice to have a "Comment posted on PR " at the end of the run.

OK for both.

Anyway while I do like the functionality provided by the script, it's a little bit too monolithic. I'd like to have a clear cut between the different components (running the benchmarks, setting up the repo, ...) because they could be easily reused separately.

antoinewdg:
You are right, let's have it merged for now (I made the changes).

> This kind of functionality could even be a nice cargo sub-command!

I have no experience with cargo subcommands, can it use python behind the scenes, or would it need a rewrite in Rust ?

Luthaf:
Yes, a cargo sub-command is anything called `cargo-<foo>` in your PATH. So you can go with Python, Ruby, Shell, Rust ...

`cargo benchcmp` started as a Python script too.

antoinewdg:
That's cool, did not know that !

PR #148, opened 2017-04-17
Adds rotation matrix as method of Matrix3.

g-bauer:
This PR moves the function `rotation_matrix` from the rotational MC move to be a method of `Matrix3` and adds a doctest.

I also modified the old function to compute the norm of `axis` inside the function.

### DIFF
--- a/src/core/src/types/matrix.rs
+++ b/src/core/src/types/matrix.rs
@@ -91,6 +91,51 @@ impl Matrix3 {
                        [m20, m21, m22]]}
     }
 
+
+    /// Returns rotation matrix given a rotation angle and an axis.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use lumol::types::{Vector3D, Matrix3};
+    /// let e1 = Vector3D::new(1.0f64, 0.0, 0.0);
+    /// let e3 = Vector3D::new(0.0f64, 0.0, 1.0);
+    /// let angle = 90f64.to_radians();
+    ///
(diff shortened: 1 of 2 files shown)


Luthaf:
They may have been a misunderstanding: the `Cargo.toml` file should stay around. The `Cargo.lock` is the one that should not be checked in the repo =)

g-bauer:
Oops! That was not a missunderstanding, I simply removed the wrong file on accident.

src/core/src/types/matrix.rs
```diff
@@ -91,6 +91,51 @@ impl Matrix3 {
...
+    /// let angle = 90f64.to_radians();
+    ///
+    /// let rotation = Matrix3::rotation_matrix(&(e1*2.0), angle);
+    /// let mut rot_vec = rotation * e3;
+    /// for i in 0..3 {
+    ///     rot_vec[i] = (rot_vec[i] * 1.0e8).round() / 1.0e8
+    /// };
+    /// assert_eq!(rot_vec, Vector3D::new(0.0, 1.0, 0.0));
+    /// ```
+    pub fn rotation_matrix(axis: &Vector3D, angle: f64) -> Matrix3 {
```
Luthaf: Should we call this function `rotation`? The matrix part is always implied now that this is a method

g-bauer: Good idea!

Luthaf:
Can you squash the commits? I'll merge after that!

g-bauer:
Done. Thank you!

Luthaf:
Thank you !

PR #154, opened 2017-04-24
Try to fix the master build issue with documentation

Luthaf:
This PR try to fix the intermittent master build failure by pulling the gh-pages branch before pushing the update.


antoinewdg:
Since you always fixed the broken builds I never saw the build log on failure, I'm actually not sure what is the issue. From what I gathered, there are some situations where there is a new commit made on `gh-pages` between the moment the build is initiated and the moment `publish-docs.sh` tries to push the new docs to the branch. Am I in the right ?

Luthaf:
Yes, you're right! The build log looks like this:

```
+git config user.name 'Travis CI'
+git config push.default simple
+git add --all .
+git diff --cached --exit-code --quiet
+git commit -m '[AUTO-COMMIT] Documentation update'
[gh-pages a2d436c] [AUTO-COMMIT] Documentation update
 60 files changed, 449 insertions(+), 130 deletions(-)
+git push
To https://<TOKEN>@github.com/lumol-org/lumol.git
 ! [rejected]        gh-pages -> gh-pages (fetch first)
error: failed to push some refs to 'https://<TOKEN>@github.com/lumol-org/lumol.git'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

Luthaf:
And I just realised that my github token is visible in this build log. I'll add a commit to this PR to change the token and add `--quiet` everywhere.

antoinewdg:
What I do not get is in what cases does this happen ? When do we have two master commits running on Travis CI at the same time ?

Luthaf:
Yes, I believe so. Both of them get the same branch, and one push faster than the other.

Luthaf:
I removed the token in `.travis.yml` to set it directly on travis website, and made @lumol-bot the author for documentation updates. This should be OK for me.

I can temporary activate the documentation build on this branch to tests the changes, what do you think?

antoinewdg:
> I can temporary activate the documentation build on this branch to tests the changes, what do you think?

Yes sure.

Does the lumol-bot have the rights to push in the repo ?

Also, if the issue is caused by two different commits on master, is rebasing the solution ? Shouldn't we just let the most recent one overwrite the other ?

Luthaf:
> Does the lumol-bot have the rights to push in the repo ?

No, good catch. Should we give him rights, or should we use another account for this?

> Also, if the issue is caused by two different commits on master, is rebasing the solution ? Shouldn't we just let the most recent one overwrite the other ?

Maybe. I wanted to play it safe here, but a git push --force might be the solution.

antoinewdg:
> Maybe. I wanted to play it safe here, but a git push --force might be the solution.

The way I see it, if the current commit is the latest one we force push, else we don't do anything. I don't understand what would happen when we rebase, does it make sense with generated documentation ?

Luthaf:
I went with the force push solution, and gave push right to the bot. The last commit is checking the code, and I'll remove it before merging this!

Luthaf:
It works! https://github.com/lumol-org/lumol/commit/072618d96b26d155b24567d919630442c84b3956. I removed the test commit, so we can merge this.

Issue #138, opened 2017-04-04
Simplify the System API?


Luthaf:
Opening a meta-issue to discuss possible simplification for the System API. The `System` is a bit of a God Object, with an ever growing API.

What we we do to simplify it? And what can we do to encapsulate the internals better, while retaining control on performances?

antoinewdg:
I don't know if this comment belongs here or in #126.
The more I think about it, the more I feel that encapsulating too many things will be a hell to do. For example, if we want to encapsulate access to particle positions, we will have to make 4 different getter functions (ref, ref mut, unsafe ref, unsafe ref mut) and I do not think we want that.

How I would see a new system is:
* a very transparent `SystemGeometry` (or `SystemConfiguration`) that contains:
  * the `UnitCell`
  * the `molecules`
  * an SOA equivalent of `Vec<Particle>`
  * the molecule ids
  * how I see it all of these fields may be public
* `Interactions` the same as before
* and the `System` object that contains:
  * the `SystemGeometry`
  * the `Interactions`
  * the `step` 
  * the `external_temperature`

There is still some weirdness with the `ParticleKinds` associations. Am I right saying that this mapping is used only for debugging/logging purposes and when building the system ?
Do we actually need the particle name -> particle kind association after building, or just the reverse one ?

Luthaf:
> how I see it all of these fields may be public

They can not really be public: we need to ensure consistency between molecules and the molecules ids list.

> Am I right saying that this mapping is used only for debugging/logging purposes and when building the system ? Do we actually need the particle name -> particle kind association after building, or just the reverse one ?

We mainly use the kind during a simulation, and the name to set the particle kind when adding a new particle. This can happen in two cases: adding new particles between simulation runs, or adding molecules during the run (GCMC is the main culprit here, and maybe Gibbs ensemble simulations too). 

So we need the name => kind association to add new particles during a simulation.
We need the kind => name association for the error messages when no interaction is found.


---


Other than that, I agree with you that having all the access functions will be bad, and the separation seems reasonable.

antoinewdg:
> They can not really be public: we need to ensure consistency between molecules and the molecules ids list.

At least the read-only part could be ?

Do we really need to add particles by name when doing that in the middle of a simulation ?

Luthaf:
[...]

Issue #126, opened 2017-03-26
Separate System and Interactions


antoinewdg:
If we want to do any parallelization, we will need the `System` object to be `Send` and `Sync`, and it is not `Sync` because of the `RefCell`s used in `Interaction`. 
The good news is that we actually do not need to have the interactions passed around threads, so making the system (here the configuration of the particles) and the interactions two separate entities would solve the problem. The bad news is that it would require some non trivial refactoring. I'll work on a draft for this and open a PR. 

As a side note : according to [the rust doc](https://doc.rust-lang.org/std/cell/index.html#introducing-mutability-inside-of-something-immutable), it would make more sense if `GlobalPotential::energy` (and others) were immutable and the `RefCell` was actually inside the struct implementing `GlobalPotential` if it were to use caching, the `System` should not be aware of this mutability. Is there a particular reason that it is done the other way around ?

antoinewdg:
Would having a separation like that be OK ?

```rust
#[derive(Clone)]
pub struct SystemConfiguration {
    /// Unit cell of the system
    cell: UnitCell,
    /// List of particles in the system
    particles: Vec<Particle>,
    /// Molecules in the system
    molecules: Vec<Molecule>,
    /// Molecules indexes for all the particles
    molids: Vec<usize>,
}
/// The System type hold all the data about a simulated system.
///
/// This data contains:
///
///   - an unit cell, containing the system;
///   - a list of particles in the system;
///   - a list of molecules in the system;
///   - a list of interactions, associating particles kinds and potentials
///   - a hash map associating particles names and particles kinds.
///
/// In the implementation, the particles contained in a molecule are guaranteed
/// to be contiguous in memory. This allow for faster access when iterating over
/// molecules, and easier molecule removal in the system.
#[derive(Clone)]
pub struct System {
    configuration: SystemConfiguration,
    /// Interactions manages the associations between particles and potentials
    interactions: Interactions,
    /// Current step of the simulation
    step: u64,
    /// Externally managed temperature for the system, if the propagation
    /// algorithm does not update the velocities.
    external_temperature: Option<f64>
}
```

g-bauer:
In this context maybe it is good to talk about #18 again? Also @Luthaf  and I had a conversation about how to represent molecules, how to get bonding information from the input and how to apply a move for a certain molecule. While this is not directly connected to parallelization, it would also change the structure of `System` and maybe `Interactions`.

[...]

PR #159, opened 2017-04-25
[Discussion] Proposal for new system

antoinewdg:
Here is my proposal for how the new System API would look like.
I want to use the `ParticlesRef` and `ParticlesRefMut` structs because while they cause a little duplicate code on the system code, they make for a (in my opinion) much nicer api.
Instead of having something like `system.particles().names()[i]` we can have `system.particles().names[i]` (I really don't like parenthesis before indexing) and more importantly:
```rust
let particles = system.particles();
...
let qi = particles.charges[i];
let pi = particels.positions[i];
...
```

This is of course not a definitive proposal, just a concrete way to fuel the discussion started in #138 and #126 .

### DIFF
--- a/src/core/src/sys/new_system.rs
+++ b/src/core/src/sys/new_system.rs
@@ -1,0 +1,257 @@
+// Lumol, an extensible molecular simulation engine
+// Copyright (C) 2015-2016 Lumol's contributors — BSD license
+
+//! The base type for simulation data in `lumol` is the `System` type.
+//!
+//! An `System` consists of a list of `Particle`; a list of `Molecule`
+//! specifying how the particles are bonded together; an unit cell for boundary
+//! conditions; and the interactions between these particles.
+use std::ops::{Index, IndexMut};
+use std::slice;
+use std::cmp::{min, max};
+use std::iter::IntoIterator;
+use std::i8;
+use std::collections::BTreeMap;
+
+use energy::PairInteraction;
+use energy::{BondPotential, AnglePotential, DihedralPotential};
+use types::{Vector3D, Matrix3, Zero};
+
+use super::{Particle, ParticleKind};
+use super::Molecule;
+use super::{CONNECT_12, CONNECT_13, CONNECT_14, CONNECT_FAR};
+use super::UnitCell;
+use super::interactions::Interactions;
+use super::EnergyEvaluator;
+use super::molecules::molecule_type;
+
+
+/// A particle kind. Particles with the same name will have the same kind. This
+/// is used for faster potential lookup.
+// #[derive(Clone, Copy, Hash, PartialOrd, Ord, PartialEq, Eq, Debug)]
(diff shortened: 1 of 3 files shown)


Luthaf:
> I want to use the ParticlesRef and ParticlesRefMut structs because while they cause a little duplicate code on the system code, they make for a (in my opinion) much nicer api.

I definitively agree with this. Do you think we should put the AoS => SoA change in this refactoring too? Or could it be applied separately?

My plan was to use [soa-derive](https://github.com/lumol-org/soa-derive) for this, so if it is needed, I can move this crate higher in my TODO list.

src/core/src/sys/new_system.rs
```diff
@@ -0,0 +1,257 @@
...
+            geometry: SystemGeometry::default(),
+            interactions: Interactions::new(),
+            step: 0,
+            external_temperature: None
+        }
+    }
+}
+
+impl System {
+    pub fn from_cell(cell: UnitCell) -> System {
```
Luthaf: This should really be called `with_cell`. I believe you just reused the name from the current code, right?

antoinewdg: I did, but yes this is better.

antoinewdg:
> Do you think we should put the AoS => SoA change in this refactoring too?

I figured that since we are going to break everything anyway, it would be better to do this on the way. I does not really add much more work or complexity, while making it in a different PR would.

`soa-derive` looks neat, where did you find documentation for macros 1.1 ? I saw [this](https://cbreeden.github.io/Macros11/) and its look great, but if you have anything else I'd be happy to read it. I can help you with code reviews on this if you want :)

Luthaf:
There is not that much documentation for the macro 1.1 part of soa-derive. Most of the work is done with [syn](https://crates.io/crates/syn) for parsing the code, and [quote](https://crates.io/crates/quote) for generating new code.

antoinewdg:
The more I think about it, the more I'm inclined to separate SOA from this PR, because SOA may impact performance, and the refactoring may also. Having these two impacts isolated is most likely a good idea.

src/core/src/sys/new_system.rs
```diff
@@ -0,0 +1,257 @@
...
+    pub masses: &'a mut [f64],
+    pub charges: &'a mut [f64],
+    pub positions: &'a mut [Vector3D],
+    pub velocities: &'a mut [Vector3D],
+    pub molecule_ids: &'a mut [usize]
+}
+
+
+#[derive(Clone)]
+pub struct SystemGeometry {
```
g-bauer: Can we come up with a better name here? Maybe `Configuration` would be more suitable?

antoinewdg: Absolutely, this is just a filler name before we come up with a better one. If you think `Configuration` is better I'll go with that.

antoinewdg:
Closing in favor of #168

PR #160, opened 2017-04-26
[WIP] Implement `Rebuild` move using configurational bias. 

g-bauer:
This PR adds a MC move that allows to "rebuild molecules" to change conformation. To do this efficiently, we use a configurational bias scheme, which can be reused for other moves.

The reason I am publishing both configurational bias and the rebuild move together is that it is very hard to test CBMC without an implementation. If you feel that this is too much inside a single PR, we could change that.

## General idea

The general idea is to remove the selected molecule (or a part of the molecule) and then rebuild it by adding back a single particle after another. This process can be divided into these steps:

1. generate multiple candidate trial positions for a single particle (based on intramolecular potentials)
2. select one positions (based on external energy)
3. repeat from 1) for the next particle until the whole molecule is grown
4. generate trials for the old molecule in a similar fashion
5. accept/reject the new molecule 

## Current implementation

- First bead is unaltered, i.e. it will be at the same position before and after the rebuilding.
- Generates trial orientations using `Harmonic` potentials for bonds and angles 
- To check if an angle or bond potential is a `Harmonic` potential, implements `as_any` function for `BondPotential` and `AnglePotential` (can do that in a separate PR if you wish)

## Work packages

Basic:
- [x] Implement basic functionality of CBMC
- [x] Implement basic rebuild move
- [ ] Write tests for uncharged molecules (find references in literature)
- [ ] Add electrostatic interactions.

Efficiency:
- [ ] Add function for `cache` (for the moment reuse existing functions)

Refactoring:
- [ ] Maybe build a struct to keep track of the rebuilding process for CBMC?
- [ ] Refactor `prepare`  

Expose to Lumol:
- [ ] Expose move to `input` crate.
- [ ] Write documentation.

## Notes for future features building on top of this PR

**The current implementation does not work for branched molecules.** It can be expanded to deal with branches.

Must haves:
- [ ] Trial generation for arbitrary bond and angle potentials
- [ ] Trial generation for fixed bonds, angles, dihedrals
- [ ] Also insert first particle (needed for insertion/deletion)

Nice to have:
- [ ] Add early rejection scheme
- [ ] Implement dual cutoff scheme (need tracking of molecule center of mass and "radius")
- [ ] Use "cheap" potential for Rosenbluth weights (connected to dual cutoff)

Advanced:
- [ ] Recoil growth

## Final comments

There is a lot of stuff that is WIP - the only thing I'm sure of at the moment is that it compiles and the tests for CBMC are working. Actually, it's hard to test stochastic code and the way I'm doing it here may be far from optimal, but hey, it's something. If you have ideas for good tests, please let me know.

It may also make sense to structure the whole CBMC stuff differently using `Struct`s instead of just functions or adding `Enums` to handle all the `match`es. We will reuse it for insertion and deletion so it should be really convenient to use.

As always, I'm open for ideas from your side concerning refactoring, style improvement, critical questions and so on. Thanks for reading!

### DIFF
--- a/src/core/src/energy/functions.rs
+++ b/src/core/src/energy/functions.rs
@@ -1,6 +1,7 @@
 // Lumol, an extensible molecular simulation engine
 // Copyright (C) 2015-2016 Lumol's contributors — BSD license
 
+use std::any::Any;
 use energy::Potential;
 use energy::{PairPotential, BondPotential, AnglePotential, DihedralPotential};
 
@@ -34,8 +35,12 @@ impl PairPotential for NullPotential {
     fn tail_energy(&self, _: f64) -> f64 {0.0}
     fn tail_virial(&self, _: f64) -> f64 {0.0}
 }
-impl BondPotential for NullPotential {}
-impl AnglePotential for NullPotential {}
+impl BondPotential for NullPotential {
+    fn as_any(&self) -> &Any { self }
+}
+impl AnglePotential for NullPotential {
+    fn as_any(&self) -> &Any { self }
+}
 impl DihedralPotential for NullPotential {}
 
 /// Lennard-Jones potential.
@@ -143,8 +148,12 @@ impl PairPotential for Harmonic {
     fn tail_virial(&self, _: f64) -> f64 {0.0}
 }
 
-impl BondPotential for Harmonic {}
-impl AnglePotential for Harmonic {}
+impl BondPotential for Harmonic {
+    fn as_any(&self) -> &Any { self }
+}
+impl AnglePotential for Harmonic {
+    fn as_any(&self) -> &Any { self }
+}
 impl DihedralPotential for Harmonic {}
 
 /// Cosine harmonic potential.
@@ -194,7 +203,9 @@ impl Potential for CosineHarmonic {
     }
 }
 
-impl AnglePotential for CosineHarmonic {}
+impl AnglePotential for CosineHarmonic {
+    fn as_any(&self) -> &Any { self }
+}
 impl DihedralPotential for CosineHarmonic {}
 
 /// Torsion potential.
--- a/src/core/src/energy/mod.rs
+++ b/src/core/src/energy/mod.rs
@@ -59,6 +59,7 @@
 //! [DihedralPotential]: trait.DihedralPotential.html
 //! [GlobalPotential]: trait.GlobalPotential.html
 //! [CoulombicPotential]: trait.CoulombicPotential.html
+use std::any::Any;
 use types::{Matrix3, Vector3D};
 
 /// A potential for force and energy computations.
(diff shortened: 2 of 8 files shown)


g-bauer:
Also, this PR is a means to compete with @antoinewdg 's PR-spree. :rofl:

Luthaf:
Wow, nice PR intro! Do you want us to review the code now, or later, when more of it is implemented?

> The current implementation does not work for branched molecules. It can be expanded to deal with branches.

Do you plan to add this feature? If so, it may be better to add it in a separated PR, and have a check for now that the molecule is linear.

Luthaf:
> To check if an angle or bond potential is a Harmonic potential, implements as_any function for BondPotential and AnglePotential (can do that in a separate PR if you wish)

Have this in the PR is OK, but it should be in a separated commit.

g-bauer:
> Do you want us to review the code now, or later, when more of it is implemented?

I guess there will be bugs in the code, especially in the `preparation` part. I'll try to write a small test simulation to make sure that the move works and that may involve some rewriting. I'd say it is better if we wait with reviews until we are sure that the basic functionality is working correctly.

 > Have this in the PR is OK, but it should be in a separated commit if possible.

Mhm...if it is ok for you, I'd favor doing a separate PR then. I'm always scared if I have to rewrite my git history.

> Do you plan to add this feature? If so, it may be better to add it in a separated PR, and have a check for now that the molecule is linear.

In the long run I think it's mandatory to have but it's another big package and I didn't dig into the implementation details yet. For now, we could add checking if there are branches, sure.

antoinewdg:
> I'm always scared if I have to rewrite my git history.

You're not alone :smile: .

Looks neat! What does "generate trials for the old molecule in a similar fashion" mean in the `General idea` section ?

g-bauer:
> What does "generate trials for the old molecule in a similar fashion" mean in the General idea section ?

To explain, I need to go a bit more into detail. 

When building a new molecule, we are inserting particles one after another. Beforehand, we don't know what the final (new) configuration will look like, right? To increase the chances, that the configuration we are building actually "makes sense" (has a good probability to be accepted), we introduce a biasing scheme, but we have to somehow account for introducing this bias.

So we use a "similar" strategy for the old molecule (that is, the selected molecule that we are going to rebuild). The difference is that we are already knowing what the configuration looks like (it's just the current positions of the atoms of the selected molecule). With "similar fashion" I mean that we compute a bias similar to the building move except that there are two differences: First, we skip the selection of positions - since we already know them (it's just the old positions). And second, we reduce the number of configurational bias steps by one and include the weight of the old positions as additional step. That way, we "retrace" the old molecules' position and can compute the "weights" that are then used for the acceptance criterion.

TLDR: We introduce a bias to build the molecule so that it will be accepted more often. To account for the bias, we use the same method for the old molecules' configuration with the difference that positions don't have to be selected.

In the `preparation` function for `Rebuild`, you will see that the rebuilding loop is `for step in 0..self.cb_steps` and there is a `select_position` call for every atom. In contrast, the retracing loop for the old molecule runs over `for step in 1..self.cb_steps` and there is no selection, the positions of the "old molecule" are used.

PR #156, opened 2017-04-25
Parallelize wolf energy and virial

antoinewdg:
This PR introduces simple parallelization of energy and virial computation in the Wolf summation method.

PR #161, opened 2017-04-27
Implement parallel shortcuts

antoinewdg:
As suggested in #156, this PR introduces a trait that enables a shorter version of `into_par_iter().map()`.

I did not know where to put it so I put it in `types`, I don't think it belongs here.


Luthaf:
I think we can put this in the `utils` module. How about calling the trait `ParrallelMap`?

antoinewdg:
> How about calling the trait ParrallelMap?

With the current name we could add other shortcuts as `par_for_each` if needed.

The `utils` module is private, is that an issue ?

Luthaf:
> The utils module is private, is that an issue ?

I see this trait as "for internal use" mainly, so this is fine. If it is useful to others, we can try to add it to rayon directly.

antoinewdg:
I moved the trait to `utils`. What's your final take on the name ?
About moving it to rayon, I don't know if this is something every user of rayon would want.

src/core/src/utils/parallel_shortcuts.rs
```diff
@@ -0,0 +1,32 @@
+use std::ops::Range;
+
+use rayon::prelude::*;
+use rayon::iter::{Map, MapFn};
+
+
+/// Utility trait that adds shortcuts for
```
Luthaf: Can you put the doc on one line?

src/core/src/utils/parallel_shortcuts.rs
```diff
@@ -0,0 +1,32 @@
...
+pub trait ParallelShortcuts : IntoParallelIterator + Sized {
+
+    /// Shortcut for `into_par_iter().map()`
+    fn par_map<F, R>(self, map_op: F) -> Map<Self::Iter, MapFn<F>>
+        where F: Fn(Self::Item) -> R + Sync, R: Send{
+        self.into_par_iter().map(map_op)
+    }
+}
+
+impl<T> ParallelShortcuts for Range<T>
```
Luthaf: Can we implement ParallelShortcuts for T: IntoParallelIterator?

antoinewdg:
Done :)

PR #162, opened 2017-05-02
Add tests for Wolf MC caching

Luthaf:


antoinewdg:
I really like the looks of `system_from_xyz` for small testing systems.
OK for me, @g-bauer you said you would take a look so I'll let you hit the button :)

g-bauer:
Could you explain why we need `let check = wolf.clone();` here instead of directly using `wolf`?

Luthaf:
> Could you explain why we need let check = wolf.clone(); here instead of directly using wolf?

Because we want to test the code without knowing the internals. As far as the test code is concerned, Wolf could cache some data, and using a separated instance alllow to be certain that we don't rely on this cached data for the `exepected` value.

Also, it is a copy/paste from Ewald! ^_^

PR #166, opened 2017-05-03
Use a vector instead of a map for composition

Luthaf:
We get O(1) composition retrivial, and no branching in energy and virial computation.

Also, the same bug as in #164 was present for virial: https://github.com/lumol-org/lumol/commit/88d3e9943ff78e14b37679007e2a410dc7e79ca0#diff-e3aff239f89d25b68f9bc977e87d49edL199

### DIFF
--- a/src/core/src/sys/composition.rs
+++ b/src/core/src/sys/composition.rs
@@ -1,0 +1,117 @@
+// Lumol, an extensible molecular simulation engine
+// Copyright (C) 2015-2016 Lumol's contributors — BSD license
+
+use std::ops::{Index, IndexMut};
+use sys::ParticleKind;
+
+/// The system composition contains the number of particles of each kind
+/// in the system.
+///
+/// # Examples
+/// ```
+/// # use lumol::sys::{Composition, ParticleKind};
+/// let mut composition = Composition::new();
+/// composition.resize(10);
+///
+/// composition[ParticleKind(2)] = 56;
+/// composition[ParticleKind(8)] = 2;
+/// composition[ParticleKind(5)] = 42;
+///
+/// assert_eq!(composition[ParticleKind(2)], 56);
+/// assert_eq!(composition[ParticleKind(8)], 2);
+/// assert_eq!(composition[ParticleKind(5)], 42);
+/// ```
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Composition(Vec<usize>);
+
+impl Composition {
+    /// Create a new empty composition
+    ///
+    /// # Examples
+    /// ```
+    /// # use lumol::sys::Composition;
+    /// let composition = Composition::new();
+    /// assert_eq!(composition.len(), 0);
+    /// ```
+    pub fn new() -> Composition {
+        Composition(Vec::new())
+    }
+
+    /// Get the size of the composition, *i.e.* the number of different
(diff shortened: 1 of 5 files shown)


antoinewdg:
While we are at it, shouldn't we change the Kind -> name mapping in `ParticleKinds` to a `Vec` ?

Luthaf:
I started working on this as part of the System/Interaction split refactor (https://github.com/lumol-org/lumol/commit/2c3de3ed4cfb43baa048d8b57128d7659cea68b5#diff-5b43a260f73f87a50f76bdba02cca4e5R14 in the refactor branch), so I prefer to land it later =)

antoinewdg:
OK, then, same as the other one I'll let @g-bauer press the button.

src/core/src/sys/composition.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// ```
+    /// # use lumol::sys::Composition;
+    /// let composition = Composition::new();
+    /// assert_eq!(composition.len(), 0);
+    /// ```
+    pub fn new() -> Composition {
+        Composition(Vec::new())
+    }
+
+    /// Get the size of the composition, *i.e.* the number of particles kind
```
g-bauer: Do you mean "the number of different particle kinds" here?

src/core/src/sys/composition.rs
```diff
@@ -0,0 +1,117 @@
...
+    ///
+    /// composition.resize(10);
+    /// assert_eq!(composition.len(), 10);
+    /// ```
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Resize the composition to hold `size` items. The new particles kinds
+    /// starts with no associated particles.
```
g-bauer: start

g-bauer:
I like it. 

The only thing I find a bit weird is the call to get the number of particles `let nj = composition[j] as f64;`. When reading this line, my intuition would not be to get the number of particles. We can go with the index (it's elegant) but maybe something like `let nj = composition.number_particles_of(j)` could also work for you?

Luthaf:
> maybe something like let nj = composition.number_particles_of(j) could also work for you?

This feel a bit verbose for a simple operation. With `composition[i]`, I am getting the composition for the particle kind `i`.

g-bauer:
> This feel a bit verbose for a simple operation. With composition[i], I am getting the composition for the particle kind i.

Sure, let's go with index then. It might be hairsplitting from my side, just felt not so intuitive. I would assume to get a molar fraction or another mixture property since I think of mixtures when I read composition. But hey, that's why we have a documentation for the API. 😉

Luthaf:
Well, having a documentation is nice, but it is better if the code is intuitive (but as we are seeing it, intuition is not the same for everyone ^^). Maybe we could rename composition to something different then?

g-bauer:
I think it is ok. The documentation of the struct makes the purpose and use case very clear. My only ideas would be `ParticleCount` or `ParticleNumbers` (composition could also be connected to molecules) but I don't know if that's more declarative.

I don't want to over-engineer this issue. You shall have the last word. 😄

Luthaf:
> You shall have the last word. 😄

Not every time! 😄  Input from different people if very valuable, specifically when dealing with API understandability.

Anyway, we can merge this, and improve it later!

PR #167, opened 2017-05-06
Remove the date from the copyright header

Luthaf:
As @g-bauer pointed out somewhere, it was out of date, and will be a hurdle to maintain. So let's get rid of it =)


antoinewdg:
Best, code review, ever.

Luthaf:
Did you actually read everything ? If so, congratulations !

PR #168, opened 2017-05-08
Separate the system in configuration and interactions

### LINKED ISSUE #126: Separate System and Interactions

### LINKED ISSUE #138: Simplify the System API?

Luthaf:
Fix #138 and fix #126.

## Status

- [x] update lumol-core 
- [x] update lumol-input
- [x] update benchmarks
- [x] update integration tests
- [x] update examples

## Structures changes

The new system look like this:

```rust
pub struct System {
    /// The system configuration
    configuration: Configuration,
    /// All the interactions in this system
    interactions: Interactions,
    /// Association particles names to particle kinds
    kinds: BTreeMap<String, ParticleKind>,
    /// The current simulation step
    step: u64,
    /// Externally managed temperature for the system
    external_temperature: Option<f64>,
}
```

`Configuration` contains the particles, molecules and unit cell. `Interaction` only works with particle kind, and does not know anything about the particles names. 

The system is now managing the name => kind associations, and the kind => name association is no longer needed (we can always access the original particle name).

## API changes

The `System` implement `Deref<Target=Configuration>`: this means that all the old code can work with minimal changes. 

The Interactions methods are mirrored by the system (`System::***_potential`), and the `Interaction` struct is now private. This reduce the API surface of the crate.

In the `Configuration`, the unit cell is public (with Deref, this means that `system.cell.volume()` works), the molecules are accessed with `system.molecules()` and the particles are accessed with indexing `system[i]`. Do you think we should also have a `system.particles()` function ? We already have `system.iter()`.

## How to read this PR

I created a `sys2` module alongside the `sys` module, and ported all that I could one file at the time, re-exporting things from sys2 to sys to keep the code compiling. The current last commit of this PR replaces the sys module by the sys2 module. Because of this, I think it will be easier to review the PR one commit at the time.

You can test the current changes locally with `cargo test -p lumol-core`, the input and integration tests will not compile.


src/core/src/sys/config/configuration.rs
```diff
@@ -0,0 +1,693 @@
...
+                    (false, i)
+                }
+            }).0, "Unsorted molecule ids {:?}", self.molids);
+        }
+
+        return delta;
+    }
+}
+
+impl<'a> IntoIterator for &'a Configuration {
```
antoinewdg: It's subjective, but I don't really like that the system (or configuration) can be iterated. Add `.particles()` in the end is not so long and much more readable.

src/core/src/energy/global/ewald.rs
```diff
@@ -752,7 +751,7 @@ impl GlobalPotential for SharedEwald {
 
     fn energy(&self, system: &System) -> f64 {
```
antoinewdg: Shouldn't these functions take the `Configuration` as argument ?

antoinewdg:
Kind of hard to review. Most of the changes make the `System` calls much simpler, so it's great :smile: 

> Do you think we should also have a system.particles() function ? We already have system.iter().

I do. As I said in the review I do not like that the system can be iterated, I also am not a fan of indexing the system.

src/core/src/energy/global/ewald.rs
Luthaf: Yep, you're right!

src/core/src/sys2/interactions.rs
```diff
@@ -0,0 +1,369 @@
...
+use std::f64;
+use std::collections::BTreeMap;
+use std::cmp::{min, max};
+
+use energy::{PairInteraction, BondPotential, AnglePotential, DihedralPotential};
+use energy::{GlobalPotential, CoulombicPotential};
+use sys::ParticleKind as Kind;
+
+/// Sort pair indexes to get a canonical representation
+#[inline] fn normalize_pair(i: Kind, j: Kind) -> (Kind, Kind) {
```
g-bauer: Name nitpicking: We could call this (and the following) functions `sort_xxx`?

g-bauer:
Great job, I really like that the whole `system.interaction.` is gone. 

> Do you think we should also have a system.particles() function ? We already have system.iter().

I'm also in favor of having a `.particles()` function, it's more declarative.

src/core/src/sys2/interactions.rs
Luthaf: Why not. It feels more like a normalization to me, but I guess this name might not be the best =)

antoinewdg: The thing is it's not always a sort, for example for an angle ABC you only sort A and C, B is always in the middle. I think it used to be called sort and it changed.

g-bauer: I was just wondering, since in every comment where these functions appear, we talk about "sorting".

Luthaf: Maybe the comments where not updated when the name changed =/ (my bad ...) I can update them now =)

Luthaf:
> I'm also in favor of having a .particles() function, it's more declarative.

I have a question for this: I used `fn particles(&self) -> std::slice::Iter<Particles>`, *i.e.* a function returning an iterator directly. This allow to write

```rust
for (i, part) in system.particles().enumerate() {

}
```

W could also have `fn particles(&self) -> &[Particles]`, which make it easier to understand just by looking at the docs. But then we need to write (with an additional `iter`)

```rust
for (i, part) in system.particles().iter().enumerate() {

}
```

Which one do you prefer?

Luthaf:
> I also am not a fan of indexing the system.

I'd rather keep the indexing functionalities. Removing it means having `Configuration::particle` and `Configuration::particle_mut` in addition to `Configuration::particles` and `Configuration::particles_mut`. And we'll need to replace code like this:

```rust
system[0].position = Vector3D::new(3, 4, 5)
```

With this:
```rust
{
    let particles = system.particle_mut(0);
    particle.position = Vector3D::new(3, 4, 5)
}
```

Also, just removing the impl gives me 180 errors =)

antoinewdg:
> I'd rather keep the indexing functionalities. Removing it means having Configuration::particle and Configuration::particle_mut in addition to Configuration::particles and Configuration::particles_mut. And we'll need to replace code like this:

Wouldn't `Configuration::particles` and `Configuration::particles_mut` be enough ?
```rust
system.particles_mut()[0].position = Vector3D::new(3, 4, 5);
```

> Also, just removing the impl gives me 180 errors =)

This is going to happen when moving to SOA anyways :).

g-bauer:
> Which one do you prefer?

I prefer the iterator `fn particles(&self) -> std::slice::Iter<Particles>` version.

antoinewdg:
I thought we did not use iterators that much because they were not on par performance wise ?

Luthaf:
> I thought we did not use iterators that much because they were not on par performance wise ?

I only recall this for pairs iterators (like `for (i, j) in system.pairs()`), not for direct linear access. Here, the function is only doing a shortcut `system.particles().iter()` === `system.particles()`.

> `system.particles_mut()[0].position = Vector3D::new(3, 4, 5);`

Yeah, you're right. I though you could not assign to a function call, but it looks like you can!

> This is going to happen when moving to SOA anyways :).

=/ I forgot about this. Yes, we'll need to move away from indexing anyway.

Luthaf:
I removed indexing. I am not 100% satisfied of the result, particularly with code like this:

```rust
for (i, pi) in system.molecule(self.molid).iter().enumerate() {
    system.particle_mut(pi).position = self.newpos[i];
}
```

But I think this can be solved by having a `MoleculeRef<'a>` type that contains a slice of particles (as proposed in https://github.com/lumol-org/lumol/issues/18#issuecomment-268860305, and then writing code like this:

```rust
for (i, particle) in system.molecule_mut(self.molid).particles_mut().enumerate() {
    particle.position = self.newpos[i];
}
```

But this will be another PR, this one is already too big =)

antoinewdg:
What's left to do here (except merge conflicts) ?

Luthaf:
Nothing on my side =). I'll rebase this and then we can run some benchmarks just to check!

Luthaf:
Ping? The build staled 2 time, but restarting it made the error go away.

Luthaf:
If you don't have more comments, I'll merge this PR tomorrow!

antoinewdg:
It's OK for me, sorry for the delay

src/core/src/energy/global/ewald.rs
antoinewdg: Is there a reason this has not be done here, or did you just forget ? I can make a PR for this, this way we should be able to completely remove the lock in Ewald :)

Luthaf: I forgot, sorry. If you have time for a PR, please do!

PR #169, opened 2017-05-10
Implement ParallelShortcuts for ndarray::Zip

antoinewdg:
The previous implementation did not work because `ndarray::Zip` does not implement `rayon::IntoParallelIterator` but a custom `ndarray_parallel::NdarrayIntoParallelIterator`, this PR changes that.

Note that we had to roll back to implementing `Parallelhortcuts` only for `Range<T>` and not for all `T` that implements `IntoParallelIterator`.

### DIFF
--- a/src/core/src/parallel/shortcuts.rs
+++ b/src/core/src/parallel/shortcuts.rs
@@ -1,8 +1,12 @@
 // Lumol, an extensible molecular simulation engine
 // Copyright (C) Lumol's contributors — BSD license
 
-use rayon::prelude::*;
+use std::ops::Range;
+
+use rayon::prelude::{ParallelIterator, IntoParallelIterator};
 use rayon::iter::{Map, MapFn};
+use ndarray_parallel::NdarrayIntoParallelIterator;
+use ndarray::Zip;
 
 
 /// Utility trait that adds shortcuts for `IntoParallelIterator` structs.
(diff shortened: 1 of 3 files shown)


antoinewdg:
No we get a compiler error:

```
   Compiling lumol-core v0.1.0 (file:///home/antoinewdg/Development/lumol/src/core)
error[E0119]: conflicting implementations of trait `parallel::shortcuts::ParallelShortcuts` for type `ndarray::Zip<_, _>`:
  --> src/parallel/shortcuts.rs:45:1
   |
36 |   impl<T> ParallelShortcuts for T
   |  _- starting here...
37 | |     where T: IntoParallelIterator {
38 | |     type Iter = <T as IntoParallelIterator>::Iter;
39 | |
40 | |     fn _into_par_iter(self) -> Self::Iter {
41 | |         self.into_par_iter()
42 | |     }
43 | | }
   | |_- ...ending here: first implementation here
44 | 
45 |   impl<Parts, D> ParallelShortcuts for Zip<Parts, D>
   |  _^ starting here...
46 | |     where Zip<Parts, D>: NdarrayIntoParallelIterator {
47 | |     type Iter = <Zip<Parts, D> as NdarrayIntoParallelIterator>::Iter;
48 | |
49 | |     fn _into_par_iter(self) -> Self::Iter {
50 | |         self.into_par_iter()
51 | |     }
52 | | }
   | |_^ ...ending here: conflicting implementation for `ndarray::Zip<_, _>`

```

antoinewdg:
(is this OK to merge ?)

Luthaf:
I am not completely satisfied by the code, let me try something for 1h, and then we can merge this =)

src/core/src/parallel/shortcuts.rs
```diff
@@ -15,13 +19,34 @@ use rayon::iter::{Map, MapFn};
 /// let s = (0..100_i32).par_map(|i| -i).sum();
 /// assert_eq!(-4950, s);
 /// ```
-pub trait ParallelShortcuts : IntoParallelIterator + Sized {
+pub trait ParallelShortcuts: Sized {
+    /// The iterator type
+    type Iter: ParallelIterator;
+
+    /// Dummy function to make it work
+    fn _into_par_iter(self) -> Self::Iter;
```
Luthaf: Can we implement `par_map` directly instead of using a dummy function?

Luthaf:
I could not find a better solution, so let's settle for this one! Sorry for the delay here.

src/core/src/parallel/shortcuts.rs
antoinewdg: We can but if we add other functions we will have to implement them all, while here we only need this one.

Luthaf: I try to follow the YAGNI principle here: only add functionalities when you need it, not before =)

antoinewdg: And I agree 100% with that, but now that it's there is seems a little extreme to remove it knowing that we will use it later, just for the sake of it. No ? :)

Luthaf: We don't know if we will use it later, that's what YAGNI is about ;-). And it feels wrong to have a "dummy" function in a trait, just for the sake of it too =)

antoinewdg: I'm really not convinced, but this is clearly not worth fighting over so I made the changes.

Luthaf: Thank you for making this change even if we disagree on this matter =)

PR #178, opened 2017-06-06
Use the builder pattern for trajectories

Luthaf:
A bit of cleanup that allow to expose all the `chemfiles::Trajectory` options to users of lumol.

### DIFF
--- a/src/core/src/sys/system.rs
+++ b/src/core/src/sys/system.rs
@@ -51,7 +51,7 @@ impl System {
         System::with_configuration(configuration)
     }
 
-    /// Create a system with the specified `configuration`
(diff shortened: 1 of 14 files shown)


src/core/src/sys/system.rs
```diff
@@ -117,6 +117,26 @@ impl System {
         }
         self.external_temperature = temperature;
     }
+
+
+
+    /// Guess the bonds in the configuration using the chemfiles algorithm.
+    ///
+    /// This function remove any existing bond, and try to guess them using a
```
antoinewdg: removes, tries

Luthaf:
Done!

PR #179, opened 2017-06-07
Speedup Ewald::density_fft

antoinewdg:
Simple change to avoid multiple access to a mutable reference and
instead use a stack value.
Improves the performance of density_fft by 40%.

PR #180, opened 2017-06-07
Do not use color codes in commits sha

Luthaf:
This was the reason the benchmark output was broken for me. The new script was used for https://github.com/lumol-org/lumol/pull/179#issuecomment-306816152.

### DIFF
--- a/scripts/benchmark-pr.py
+++ b/scripts/benchmark-pr.py
@@ -72,13 +72,6 @@ def request_api(endpoint, data=None):
     return r.json()
 
 
-def get_master_commit():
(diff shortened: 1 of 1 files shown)


antoinewdg:
Should we start using the [lib I made](https://github.com/antoinewdg/rbu) (I mentioned it in the chat) instead of the script ? If so, could you port these changes to the repo ?

Luthaf:
Yeah, why not =) We can still merge this one, and I'll try to find the time to send you a PR!

antoinewdg:
OK :)

PR #181, opened 2017-06-08
Make GlobalPotential and GlobalCache use Configuration instead of System

antoinewdg:


antoinewdg:
The final goal of this would be to remove the `RwLock` in Ewald, but this is not possible yet because `System` still needs to be `Sync` as it is passed around every thread during the regular energy, force and virial computations (for example [here](https://github.com/lumol-org/lumol/blob/fac0d20b86b98d815043da5bcb3c75c3013f8953/src/core/src/sys/compute.rs#L44)).

The solution would involve a little modification of the current structure, so I think this is better to do it in another PR, I would be happy to hear your thought about it.

Luthaf:
> The solution would involve a little modification of the current structure, so I think this is better to do it in another PR, I would be happy to hear your thought about it.

I am OK with more changes in another PR! What modifications do you have in mind?

PR #182, opened 2017-06-08
Remove a few TODO from the code

Luthaf:
Now that rust 1.18 is out, we can use restricted privacy (`pub(crate)`), and remove a few todos from the code.


Luthaf:
While I was here, I removed most of the TODO in the code. We only have one left, which I'll remove in a separated PR.

PR #189, opened 2017-06-11
Remove Ewald RwLock

antoinewdg:
This PR:
* introduces a separation between `Interactions` and `LocalInteractions`, the latter containing only the pair, bond, angle and dihedral potentials (and thus being `Sync`)
* adapts the `System` to this change
* removes the `RwLock` on `Ewald`

Maybe the `Ewald`/`SharedEwald` separation does not make that much sense anymore ? Should I remove it and only hide `rho` and `fourier_phases` behind `RefCell`s ?

### DIFF
--- a/src/core/src/sys/energy.rs
+++ b/src/core/src/sys/energy.rs
@@ -7,8 +7,9 @@
 //! potential energy of an `System`.
 
 use std::f64::consts::PI;
+use std::ops::Deref;
 
-use sys::System;
+use sys::{System, Configuration, LocalPotentials};
 use parallel::prelude::*;
 
 /// An helper struct to evaluate energy components of a system.
@@ -29,10 +30,10 @@ impl<'a> EnergyEvaluator<'a> {
     /// Compute the energy associated with the pair of particles `i, j` at
     /// distance `r`
     #[inline]
-    pub fn pair(&self, r: f64, i: usize, j: usize) -> f64 {
-        let distance = self.system.bond_distance(i, j);
+    pub fn pair(configuration: &Configuration, potentials: LocalPotentials, r: f64, i: usize, j: usize) -> f64 {
+        let distance = configuration.bond_distance(i, j);
         let mut energy = 0.0;
-        for potential in self.system.pair_potentials(i, j) {
+        for potential in potentials.pairs(i, j) {
             let info = potential.restriction().information(distance);
             if !info.excluded {
                 energy += info.scaling * potential.energy(r);
@@ -44,12 +45,15 @@ impl<'a> EnergyEvaluator<'a> {
     /// Compute the energy of all the pairs in the system
     pub fn pairs(&self) -> f64 {
 
-        (0..self.system.size()).par_map(|i| {
+        let configuration = self.system.deref();
+        let local_potentials = self.system.local_potentials();
+
+        (0..configuration.size()).par_map(|i| {
             let mut local_energy = 0.0;
 
-            for j in (i+1)..self.system.size() {
-                let r = self.system.nearest_image(i, j).norm();
-                local_energy += self.pair(r, i, j);
+            for j in (i+1)..configuration.size() {
+                let r = configuration.nearest_image(i, j).norm();
+                local_energy += Self::pair(configuration, local_potentials, r, i, j);
             }
             local_energy
         }).sum()
@@ -68,7 +72,7 @@ impl<'a> EnergyEvaluator<'a> {
             let ni = composition[i] as f64;
             for j in self.system.particle_kinds() {
                 let nj = composition[j] as f64;
-                for potential in self.system.interactions().pairs(i, j) {
+                for potential in self.system.interactions().local.pairs(i, j) {
                     energy += 2.0 * PI * ni * nj * potential.tail_energy() / volume;
                 }
             }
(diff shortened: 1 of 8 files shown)


Luthaf:
I feel like this is adding a bit of complexity, and I am not sure I get the goal of these changes. Why do you want to remove the `RwLock`?

antoinewdg:
Well it removes the requirement for `GlobalPotential` to be `Sync`, which is something that we added to not be blocked on starting parallelization. The `GlobalPotentails` actually do not need to be shared among threads, only the bonds, pairs, angles and dihedrals interactions need to be (if there is some parallelization involved in a `GlobalPotential`, it's handled by the potential itself).

Luthaf:
For Travis, I have no idea. Closing and reopening the PR triggered a build.

Luthaf:
OK, I think I get the rational. I'll try to give the code a look when I find the time !

src/core/src/sys/interactions.rs
```diff
@@ -48,11 +48,8 @@ type BondKind = (Kind, Kind);
 type AngleKind = (Kind, Kind, Kind);
 type DihedralKind = (Kind, Kind, Kind, Kind);
 
-/// The `Interaction` type hold all data about the potentials in the system.
-///
-/// Its main role is to store and provide access
 #[derive(Clone)]
-pub struct Interactions {
+pub struct LocalInteractions {
```
Luthaf: This is missing documentation

src/core/src/sys/interactions.rs
```diff
@@ -61,22 +58,28 @@ pub struct Interactions {
...
     dihedrals: BTreeMap<DihedralKind, Vec<Box<DihedralPotential>>>,
+}
+
+/// The `Interaction` type hold all data about the potentials in the system.
+///
+/// Its main role is to store and provide access
+#[derive(Clone)]
+pub struct Interactions {
+    /// Local intreactions
+    pub local: LocalInteractions,
```
Luthaf: Instead of having a pub field, how about implementing `Deref<Target=LocalInteractions>`?

src/core/src/sys/system.rs
```diff
@@ -38,6 +38,13 @@ pub struct System {
     external_temperature: Option<f64>,
 }
 
+
+/// Struct that can be used to query the
+/// local interactions. As opposed to the `System`,
+/// it is `Sync`.
+#[derive(Copy, Clone)]
+pub struct LocalPotentials<'a>(&'a Configuration, &'a LocalInteractions);
```
Luthaf: I don't really like the tuple struct here, partly because accessing the fields through `self.0` and `self.1` make the code harder to read. Also, we might not need to expose this in the public API. See my main comment for an alternative proposal.

src/core/src/sys/energy.rs
```diff
@@ -29,10 +30,10 @@ impl<'a> EnergyEvaluator<'a> {
     /// Compute the energy associated with the pair of particles `i, j` at
     /// distance `r`
     #[inline]
-    pub fn pair(&self, r: f64, i: usize, j: usize) -> f64 {
-        let distance = self.system.bond_distance(i, j);
+    pub fn pair(configuration: &Configuration, potentials: LocalPotentials, r: f64, i: usize, j: usize) -> f64 {
```
Luthaf: Why do we need to pass both a Configuration and LocalPotentials, which contains a configuration?

Luthaf:
> Maybe the Ewald/SharedEwald separation does not make that much sense anymore ? Should I remove it and only hide rho and fourier_phases behind RefCells ?

The code used to look like this (was removed in https://github.com/lumol-org/lumol/commit/2a43eaf6f9eb923c3ff148ab17bd5298703fd068), and this was a massive perf hit, due to RefCell counter increment and decrement in each function. Maybe it will be better now with all the improvements to Ewald, I don't know.

antoinewdg:
> This have the advantage of not polluting the API of System, while allowing parallel computation of the energy. What do you think?

The goal of the `LocalPotentials` is mainly to avoid the duplication of the logic of mapping a particle index to a `Kind`, how does your solution avoid that? I may not understand well, but how I see it we would also need to make a `Local` version of `Forces` and `Virial`.

> The code used to look like this (was removed in 2a43eaf), and this was a massive perf hit, due to RefCell counter increment and decrement in each function. 

In this case let's leave it this way :)

src/core/src/sys/interactions.rs
antoinewdg: I'm not a fan of abusing `Deref`, but why not.

src/core/src/sys/energy.rs
antoinewdg: Because I did not consider the configuration to be part of the public API for `LocalPotentials`, I agree it's strange.

Luthaf:
> The goal of the LocalPotentials is mainly to avoid the duplication of the logic of mapping a particle index to a Kind, how does your solution avoid that?

I did not though of this =/

antoinewdg:
After a little time to think about this, I tend to agree that this PR does not bring that much to the table considering the API changes, so if you decide it's not worth it I'll to not cry too much :stuck_out_tongue: .

Luthaf:
I am not sure what to think here. On one hand, not requiring GlobalPotential to be Sync seems worthwhile, and will remove the cost of locking a RwLock/Mutex. On the other hand, this is making the API significantly more complex, and the cost is only paid once for every energy computation.

Maybe we can open an issue to try to think about this, and link to this proposition?

Luthaf:
Ok, so let's close this, and continue thinking how we could remove RwLock in a separated issue =)

Thanks again @antoinewdg for the proposition!

antoinewdg:
Oh I thought I answered your previous comment, 100% agree :)

PR #190, opened 2017-06-13
Fix clippy warnings

Luthaf:


examples/mc_npt_ethane.rs
```diff
@@ -41,7 +41,8 @@ fn get_system() -> System {
     system.add_bond_potential("C", "C", bond);
 
     // Check if bonds are guessed correctly.
-    assert_eq!(system.size() as f64 / system.molecules().len() as f64, 2.0);
+    assert_eq!(system.size() / system.molecules().len(), 2);
```
antoinewdg: Why not simply `assert_eq!(system.size(), 2 * system.molecules().len())` ?

examples/mc_npt_spce.rs
```diff
@@ -59,9 +59,9 @@ fn get_system() -> System {
     system.set_coulomb_potential(Box::new(SharedEwald::new(ewald)));
 
     // Check if bonds are guessed correctly.
-    assert_eq!(system.size() as f64 / system.molecules().len() as f64, 3.0);
-
-    system
+    assert_eq!(system.size() / system.molecules().len(), 3);
```
antoinewdg: Same question

tests/md-nacl.rs
```diff
@@ -1,7 +1,7 @@
 // Lumol, an extensible molecular simulation engine
 // Copyright (C) Lumol's contributors — BSD license
 
-//! Testing physical properties of a NaCl crystal
+//! Testing physical properties of a sodium chloride crystal
```
antoinewdg: Is that really better ? I mean I get NaCl, sodium chloride I have to think about it ^^

src/input/src/simulations/logging.rs
```diff
@@ -31,12 +31,12 @@ impl Encode for LogEncoder {
             LogLevel::Debug => try!(write!(out, "[debug] ")),
             LogLevel::Info => {},
             LogLevel::Warn => {
-                try!(out.set_style(&Style::new().text(Color::Red)));
+                try!(out.set_style(Style::new().text(Color::Red)));
```
antoinewdg: I don't get what happens here, you can just substitute a reference for a plain struct ?

tests/md-nacl.rs
Luthaf: NaCl trigger a lint (it looks like a type), and give a suggestion to put it inside backticks ...

src/input/src/simulations/logging.rs
Luthaf: Yes, because `text` returns a reference to self. This was sending a warning about borrowing that is insta-deref by the the compiler

antoinewdg: OK :)

Luthaf:
Changes done!

PR #192, opened 2017-06-27
Morse Potential implementation

EloD10:
So sorry guys, i had to recreate the same PR because I re-fork the project : PR#191

I have listened your feedbacks and i think it's now correct =)

### DIFF
--- a/doc/src/input/potentials.md
+++ b/doc/src/input/potentials.md
@@ -129,3 +129,26 @@ integer.
 atoms = ["C", "C", "C", "C"]
 torsion = {k = "40 kJ/mol", delta = "120 deg", n: 4}
 ```
+## Morse potential
+
+This potential is usually used for intramolecular interaction such as bonds,
+angles or dihedrals. It is a better approximation for the vibrational structure
+of the molecule than the Harmonic potential. It is expressed as: $$ V(r) = depth
+(1 - \exp(- a (r - x_0))^2 $$
+
+The potential type keyword is `morse`, and the parameters `a`, `x0` and `depth`
+should be provided as strings.
+
+```toml
+[[pairs]]
+atoms = ["A", "B"]
+morse = {depth = "40 kJ/mol", a = "2.0 A^-1", x0 = "1.3 A"}
+```
+
+For angles and dihedral angles, `x0` and `a` should be provided in angle units:
+
+```toml
+[[pairs]]
+atoms = ["A", "B"]
+morse = {depth = "40 kJ/mol", a = "2.0 rad^-1", x0 = "109.7 deg"}
+```
--- a/src/core/src/energy/functions.rs
+++ b/src/core/src/energy/functions.rs
@@ -82,7 +82,7 @@ impl PairPotential for LennardJones {
         let rc3 = cutoff * cutoff * cutoff;
         let s9 = s3 * s3 * s3;
         let rc9 = rc3 * rc3 * rc3;
-        return 4.0 / 3.0 * self.epsilon * s3 * (1.0 / 3.0 * s9 / rc9 - s3 / rc3);
+        4.0 / 3.0 * self.epsilon * s3 * (1.0 / 3.0 * s9 / rc9 - s3 / rc3)
     }
 
     fn tail_virial(&self, cutoff: f64) -> f64 {
@@ -90,7 +90,7 @@ impl PairPotential for LennardJones {
         let rc3 = cutoff * cutoff * cutoff;
         let s9 = s3 * s3 * s3;
         let rc9 = rc3 * rc3 * rc3;
-        return 8.0 * self.epsilon * s3 * (2.0 / 3.0 * s9 / rc9 - s3 / rc3);
+        8.0 * self.epsilon * s3 * (2.0 / 3.0 * s9 / rc9 - s3 / rc3)
     }
 }
 
@@ -125,7 +125,7 @@ impl Potential for Harmonic {
         let dx = x - self.x0;
         0.5 * self.k * dx * dx
     }
-    
+
     fn force(&self, x: f64) -> f64 {
         self.k * (self.x0 - x)
     }
@@ -364,6 +364,53 @@ impl PairPotential for BornMayerHuggins {
     }
 }
 
+/// Morse potential
+///
+/// The following potential expression is used: `V(r) = depth * (1 - exp(a (x0 -
+/// r))^2` ; where the paramaters are 'x0' for the equilibrium value, 'depth'
+/// for the well depth, and 'a' for the well width.
+///
+/// # Examples
+///
+/// ```
+/// use lumol::energy::Potential;
+/// use lumol::energy::MorsePotential;
+///
+/// let potential = MorsePotential{a: 2.0, x0: 1.3, depth: 4.0};
+/// assert_eq!(potential.energy(1.0), 2.703517287822119);
+/// assert_eq!(potential.force(1.0), -37.12187076378477);
+/// ```
+#[derive(Clone, Copy)]
+pub struct MorsePotential {
+    /// Exponential term width value
+    pub a: f64,
+    /// Equilibrium value
+    pub x0: f64,
+    /// Well depth value
+    pub depth: f64,
+}
+
+impl Potential for MorsePotential {
+    fn energy(&self, r: f64) -> f64 {
+        let rc = 1.0 - f64::exp((self.x0 - r) * self.a);
+        self.depth * rc * rc
+    }
+
+    fn force(&self, r: f64) -> f64 {
+        let exp = f64::exp((self.x0 - r) * self.a);
+        2.0 * self.depth * (1.0 - exp * exp) * self.a
+    }
+}
+
+impl PairPotential for MorsePotential {
+    fn tail_energy(&self, _: f64) -> f64 {0.0}
+    fn tail_virial(&self, _: f64) -> f64 {0.0}
+}
+
+impl BondPotential for MorsePotential {}
+impl AnglePotential for MorsePotential {}
+impl DihedralPotential for MorsePotential {}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -486,5 +533,21 @@ mod tests {
         let e0 = born.energy(4.0);
         let e1 = born.energy(4.0 + EPS);
         assert_relative_eq!((e0 - e1) / EPS, born.force(4.0), epsilon=1e-6);
+    }
+
+  #[test]
+    fn morse() {
+        let morse = MorsePotential{a: 2.0, x0: 1.3, depth: 4.0};
+
+        // Comparing to externally computed values
+        assert_eq!(morse.energy(1.0), 2.703517287822119);
+        assert_eq!(morse.force(1.0), -37.12187076378477);
+
+        assert_eq!(morse.tail_energy(1.0), 0.0);
+        assert_eq!(morse.tail_virial(1.0), 0.0);
+
+        let e0 = morse.energy(1.3);
+        let e1 = morse.energy(1.3 + EPS);
+        assert_relative_eq!((e0 - e1) / EPS, morse.force(1.3), epsilon=1e-6);
     }
 }
--- a/src/input/src/interactions/pairs.rs
+++ b/src/input/src/interactions/pairs.rs
@@ -7,7 +7,7 @@ use lumol::units;
 
 use lumol::energy::{PairPotential, PairInteraction, BondPotential};
 use lumol::energy::{Harmonic, LennardJones, NullPotential};
-use lumol::energy::{Buckingham, BornMayerHuggins};
+use lumol::energy::{Buckingham, BornMayerHuggins, MorsePotential};
 use lumol::energy::TableComputation;
 
 use error::{Error, Result};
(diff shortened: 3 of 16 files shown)


EloD10:
WIP

EloD10:
For documentation, i don't understand exactly what i must put in .toml file ?

g-bauer:
Do you mean the ` ```toml ` part within the [potential documentation](https://github.com/lumol-org/lumol/blob/master/doc/src/input/potentials.md)?

You should provide an example of how the user can use the potential from the input, e.g. (I am making values up)
```toml
[[pairs]]
atoms = ["A", "B"]
# Additional parameters here are 'De', 're' and 'a'.
morse = {De ="40 kJ/mol", re = "3.0 A", a="12.0"}
```
I'm not sure about the units here :)

src/input/src/interactions/pairs.rs
```diff
@@ -206,6 +206,8 @@ fn read_pair_potential(pair: &Table) -> Result<Box<PairPotential>> {
             "lj" => Ok(Box::new(try!(LennardJones::from_toml(table)))),
             "buckingham" => Ok(Box::new(try!(Buckingham::from_toml(table)))),
             "born" => Ok(Box::new(try!(BornMayerHuggins::from_toml(table)))),
+            "MorsePotential" => Ok(Box::new(try!(MorsePotential::from_toml(table)))),
```
g-bauer: Since we are using short names as keys here, maybe naming it `morse` is more suitable here?

This will be the name by which the user can define the potential in the toml file, i.e.
```toml
[[pairs]]
atoms = ["A", "B"]
morse = {depth= ...
```

src/core/src/energy/functions.rs
```diff
@@ -364,6 +364,50 @@ impl PairPotential for BornMayerHuggins {
...
+/// ```
+#[derive(Clone, Copy)]
+pub struct MorsePotential {
+    /// Width value
+    pub width: f64, 
+    /// Equilibrium bond distance
+    pub r0: f64,
+    /// Well depth value
+    pub depth: f64,
+}
```
g-bauer: Some nitpicking about the names. For other potentials we used the symbols as potential parameter names, i.e. `epsilon`, `sigma` or `k`. Since you are writing the formula using `De`, `a` and `re`, would make sense to also use these as field names? This should then be consistent with what we use in the input file.

src/core/src/energy/functions.rs
```diff
@@ -364,6 +364,50 @@ impl PairPotential for BornMayerHuggins {
...
+/// The following potential expression is used: `V(r) = De * (1 - exp(-a (r - re))^2 ; where 'r' is the distance between the atoms, 
+/// 're' is the equilibrium bond distance, 'De' is the well depth, and 'a' controls the 'width' of the potential
+///
+/// # Examples
+///
+/// ```
+/// use lumol::energy::Potential;
+/// use lumol::energy::MorsePotential;
+///
+/// let potential = BornMayerHuggins{a: 2.0, r0: 3.0, De: 4.0};
```
g-bauer: Copy and paste error? BornMayerHuggins 😄

src/core/src/energy/functions.rs
```diff
@@ -487,4 +531,18 @@ mod tests {
         let e1 = born.energy(4.0 + EPS);
         assert_relative_eq!((e0 - e1) / EPS, born.force(4.0), epsilon=1e-6);
     }
+  #[test]
+    fn MorsePotential() {
```
g-bauer: Could you rename this function to something like `morse` (to be in line with `lj`, `harmonice`, etc.)? Function names should always be lower case. That may sound picky, but it makes the code more readable 😄

src/input/src/interactions/pairs.rs
EloD10: Oups ! Of course, it's done =)

src/core/src/energy/functions.rs
EloD10: Oups again ^^
(yeah, itake the same model)

src/core/src/energy/functions.rs
EloD10: In the user doc ? I hesitated to avoid losing with the Wikipedia formula: I had written the correspondence like this "` depth` ($ From $). But I'll change that!

src/core/src/energy/functions.rs
EloD10: Done !

EloD10:
mm..i'm going to study tests ^^

src/core/src/energy/functions.rs
g-bauer: I was referring to the names of the variables of the struct. My proposal would be to name the variables like the symbols in the formula, for example:
```rust
#[derive(Clone, Copy)]
pub struct MorsePotential {
    /// Width value
    pub de f64, 
    /// Equilibrium bond distance
    pub re: f64,
    /// Well depth value
    pub a f64,
}
```
I'd do that for two reasons:
1. It is in line with what we did for other potentials
2. It is easier to connect the code to the formula, since we use the same names.

This would also mean that we'd have to change the names in the `input` module.
Does that make sense?

g-bauer:
> mm..i'm going to study tests :smile:

Do you need help with the tests?

src/core/src/energy/functions.rs
EloD10: Oh, @Luthaf suggested these names so i will wait his opinion =)

EloD10:
I will read the documentation and understand more myself for now =)

src/core/src/energy/functions.rs
```diff
@@ -391,12 +391,14 @@ pub struct MorsePotential {
 
 impl Potential for MorsePotential {  
     fn energy(&self, r: f64) -> f64 { 
-        let rc2 = (1.0 - f64::exp(self.width * (r - self.r0))) * (1.0 - f64::exp(self.width * (r - self.r0)));
+        let rc = 1.0 - f64::exp(self.width * (r - self.r0));
```
g-bauer: I think you missed the minus in front of `self.width`, right?

```rust
let rc = 1.0 - f64::exp(-self.width * (r - self.r0));
```

src/core/src/energy/functions.rs
g-bauer: Sure! This was merely a proposal.

Luthaf: As long as we have the same name  everywhere (code, input, doc and user manual), I am good. I suggested changing a to width and de to depth because it is slightly easier to understand, but I don't have a strong opinion on this =)

src/core/src/energy/functions.rs
EloD10: Exact ! Thank you, i have rewrote formula.

EloD10:
I know that the tests are not good, someone have an idea of ​​the problem? (I don't think that the formula is erroneous)

src/core/src/energy/functions.rs
```diff
@@ -364,6 +364,50 @@ impl PairPotential for BornMayerHuggins {
...
+    pub r0: f64,
+    /// Well depth value
+    pub depth: f64,
+}
+
+impl Potential for MorsePotential {  
+    fn energy(&self, r: f64) -> f64 { 
+        let rc = 1.0 - f64::exp(-self.width * (r - self.r0));
+        let rc2 = rc * rc;
+        self.depth * (rc2 - 1.0)
```
Luthaf: This should be `self.depth * rc2`

src/core/src/energy/functions.rs
```diff
@@ -364,6 +364,50 @@ impl PairPotential for BornMayerHuggins {
...
+    }
+
+    fn force(&self, r: f64) -> f64 { 
+        self.depth * (f64::exp(-2.0 * self.width * (r - self.r0)) - 2.0 * f64::exp(-self.width * (r - self.r0)))
```
Luthaf: The force should be the derivative of the energy, and it does not look like it here. It calculated it and I got

```
2 * De * width * (1 - exp(-width * (r - re)) * exp(-width * (r - re))
```

Luthaf:
You should be able to run the tests locally with `cargo test - p lumol-core`, so that you don't need to push to Travis everytime =)

src/core/src/energy/functions.rs
```diff
@@ -82,15 +82,15 @@ impl PairPotential for LennardJones {
...
-        return 4.0 / 3.0 * self.epsilon * s3 * (1.0 / 3.0 * s9 / rc9 - s3 / rc3);
+        4.0 / 3.0 * self.epsilon * s3 * (1.0 / 3.0 * s9 / rc9 - s3 / rc3)
     }
 
     fn tail_virial(&self, cutoff: f64) -> f64 {
         let s3 = self.sigma * self.sigma * self.sigma;
         let rc3 = cutoff * cutoff * cutoff;
         let s9 = s3 * s3 * s3;
         let rc9 = rc3 * rc3 * rc3;
-        return 8.0 * self.epsilon * s3 * (2.0 / 3.0 * s9 / rc9 - s3 / rc3);
```
EloD10: As you see, i had just delete that keyword because it is implicit.

src/core/src/energy/functions.rs
```diff
@@ -82,15 +82,15 @@ impl PairPotential for LennardJones {
         let rc3 = cutoff * cutoff * cutoff;
         let s9 = s3 * s3 * s3;
         let rc9 = rc3 * rc3 * rc3;
-        return 4.0 / 3.0 * self.epsilon * s3 * (1.0 / 3.0 * s9 / rc9 - s3 / rc3);
```
EloD10: Same thing.

EloD10:
So sorry guys, I made several gourdes, i think that all run correctly now =)

src/core/src/energy/functions.rs
```diff
@@ -364,6 +364,50 @@ impl PairPotential for BornMayerHuggins {
...
+///
+/// # Examples
+///
+/// ```
+/// use lumol::energy::Potential;
+/// use lumol::energy::MorsePotential;
+///
+/// let potential = MorsePotential{width: 2.0, r0: 1.5, depth: 4.0};
+/// assert_eq!(morse.energy(1.0), 2.703517287822119);
+/// assert_eq!(morse.force(1.0), -37.12187076378477);
```
g-bauer: Those tests are the same as the unit tests below, right? They currently fail since *r0* is different than in the unit tests. Maybe a good additional test would be to hand the distance where force and energy are zero?

doc/src/input/potentials.md
```diff
@@ -129,3 +129,17 @@ integer.
...
+angles or dihedrals. It is a better approximation for the vibrational 
+structure of the molecule than the Harmonic potential. It is expressed as:
+$$ V(r) = De (1 - \exp(-a (r - r_e))^2 $$
+
+The potential type keyword is `morse`, and the parameters `width` ($a$),
+ `r0` ($r_e$) and `depth` ($De$) should be provided as strings.
+
+```toml
+[[pairs]]
+
```
g-bauer: Could you supply an example line for the input? For example

```toml
atoms = ["A", "B"]
# Parameters here are 'depth', 'width' and 'r0'.
morse = {depth= "40 kJ/mol", width = "3.0 A", r0 = "4.0 A"}
```
(With the right units of course)

src/core/src/energy/functions.rs
Luthaf: There is no always such distance in Morse potential, only when depth is negative. But this can be a nice addition =)

g-bauer: Ok my wording was misleading. I meant the distances (one for energy (=r0) one for force). 😄

EloD10: Done =)

src/core/src/energy/functions.rs
```diff
@@ -487,4 +531,26 @@ mod tests {
...
     }
+    
+  #[test]
+    fn morse() {
+        let morse = MorsePotential{width: 2.0, r0: 1.3, depth: 4.0};
+        
+        //http://www.wolframalpha.com/input/?i=4+*+(1+-+e%5E(-2+*+(1+-+1.3)))%5E2
+        assert_eq!(morse.energy(1.0), 2.703517287822119);
+        //http://www.wolframalpha.com/input/?i=4+*+(1+-+e%5E(-2+*+(-1+-+1.3)))%5E2
+        assert_eq!(morse.energy(-1.0), 38796.641709840165);
```
g-bauer: That's an interesting test case 😄 There shouldn't be negative distances. But I guess it tests the function.

g-bauer:
Looks good to me! 

Should we also include the input tests within this PR? 
I am wondering, is the `width` parameter something like the force constant of the harmonic potential (controlling the width)? If so, should its unit be that of an inverted length?

src/core/src/energy/functions.rs
EloD10: Yeah we can't lol ! It was just a fast test, i will remove it ^^

Luthaf:
So, with @EloD10 permission, I took over this PR to finish it.

I squashed his commits together and made a few changes, mainly reverting `width` to `a`, because a `width` in inverted length would be a bit strange as @g-bauer pointed out =)

@g-bauer, I'll let your merge this is tests passes and everything is ok for you!

----

@EloD10, as I forced pushed to your fork, you can run the following commands to update any local branch:

```
git fetch  # assuming origin is https://github.com/EloD10/lumol
git checkout dev
git reset --hard origin/dev
```

g-bauer:
Merged. :champagne: @EloD10 Thanks for the PR.

Issue #45, opened 2016-10-20
Plugin system


Luthaf:
I have been thinking about this for a while, here is a request for comment (RFC) to shape it a bit more and gather feedback. The implementation of this will take time and is not high priority, but I really want to get this right.

## Motivations

A nice way for the code to be easy to extend would be to provide a plugin system. A plugin is some piece of code that is not part of the main lumol code, and is **not compiled at the same time** than the main binary, nor linked to it.

Having a plugin system would mean that people can write their own code and use it with lumol without recompiling the whole project. It is highly desirable that this plugin system works easily with the input files too. I'd like the loading/use of a plugin to be as easy as:

```toml
[input]
version = 1

[plugins]
foo = "path/to/foo/plugin.so"

[[pairs]]
atoms = ["A", "A"]
foo = {} # potential foo is defined inside the foo plugin

[[pairs]]
atoms = ["A", "B"]
bar = {} # potential bar is also defined inside the foo plugin
```

## Types of plugins

I see three main kinds of plugins for the Lumol use case:

1. Compiled plugins written in Rust;
2. Compiled plugins written in C (or any C-exporting language: Fortran, ...);
3. Interpreted plugins written in Python.

(1) and (2) are fast, (3) is nice for writing quick prototypes. (1) is easier to implement, and (2) allow to call into other codes for potential computations (to use DFT level energy for example). (1) and (2) would be based on shared library and dynamic loading, while (3) will use some kind of future python binding.

Plugins should be usable to extend lumol in all the possibles means: not only writing new potentials, but also new integrators, new Monte-Carlo moves, ... Everything that is defined as a trait for extensibility in the code should be available as a plugin

## Proposed implementation strategy

I will focus on Rust plugins for now, as they are way easier to implement, and the C plugins can be built on the top of the Rust plugins mechanisms.

Every plugin would be compiled as a shared library, and expose at least a single function: `lumol_register`, taking a single argument of type `&mut PluginRegistry` defined in `lumol_input` crate. This function returns a `Result` in case something goes wrong. The `PluginRegistry` will define a few method to register extension points

```rust
extern crate lumol_input;
use lumol_input::{PluginRegistry, PluginError};

struct Foo;
struct Bar;

impl PairPotential for Foo {...}
impl PairPotential for Bar {...}

pub fn lumol_register(reg: &mut PluginRegistry) -> Result<(), PluginError> {
    try!(reg.add_pair::<Foo>("foo"));
[...]

PR #196, opened 2017-09-11
[WIP]Tutorial for pair potential implementation

g-bauer:
**EDIT: New structure**

This PR will teach users how to add a `PairPotential` to Lumol.

These are the steps discussed in the tutorial:
Part 1:
- [x] Introduction
- [x] Prerequisites
- [x] Building the struct
- [x] impl `Potential`
- [x] impl `PairPotential`
- [x] example simulation

Part 2 (moving the implementation into lumol core):
- [ ] add API documentation
- [ ] add unit tests
- [ ] add parser for input files
- [ ] add documentation in the manual

You can clone this PR and locally build the manual using

```
mdbook serve --open
```

from within the `lumol/doc/` directory (you can install mdbook using cargo).


doc/src/advanced/potential.md
```diff
@@ -0,0 +1,254 @@
...
+- [`DihedralPotential`](http://lumol.org/lumol/latest/lumol/energy/trait.DihedralPotential.html) for covalent dihedral angles interactions.
+
+In this tutorial, we will implement a potential to describe non-bonded pair interactions, namely the [**Mie potential**](http://www.sklogwiki.org/SklogWiki/index.php/Mie_potential). We will have to implement both the `Potential` as well as the `PairPotential` traits. If we wanted to implement a function that can be used as non-bonded as well as bond-length potential, we'd have to implement `Potential`, `PairPotential` as well as `BondPotential`. "Implementing a trait" means that we will define a [structure](https://doc.rust-lang.org/book/second-edition/ch05-00-structs.html) (a `struct`) for which we will add functions to satisfy the traits' requirements.
+
+Let's start by having a look at the API documentation for `Potential`. Open the [API documentation](http://lumol.org/lumol/latest/lumol/index.html) and navigate to the `energy` module which is listed under the `modules` section at the bottom of the page. Traits are listed at the very bottom of the module documentation. Open the documentation for the `Potential` trait.
```
Luthaf: I'll give a direct link to the trait here, instead of making the people navigate

doc/src/advanced/potential.md
```diff
@@ -0,0 +1,254 @@
...
+
+$u(x) = \frac{n}{n-m} \left(\frac{n}{m}\right)^{m/(n-m)}\epsilon \left[ \left( \frac{\sigma}{x}\right)^n - \left( \frac{\sigma}{x}\right)^m \right]$
+
+where $x$ denotes the distance between two interaction sites $i, j$, $x = x_{ij} = | \mathbf{r}_j - \mathbf{r}_i |$. The parameters of the potential are
+
+- $n, m$ the repulsive and attractive exponents, respectively,
+- $\epsilon$ the energetic paramater,
+- $\sigma$ the particle diameter or structural parameter.
+
+We start by defining the `struct` for our potential. We will add our code to the `functions.rs` file located in `lumol/src/core/src/energy`. Add the following lines:
```
Luthaf: We don't need to add new potentials in the same module as the other ones. Actually, we don't even need to add it in the same crate. See my suggestion below!

doc/src/advanced/potential.md
```diff
@@ -0,0 +1,254 @@
...
+}
+```
+
+Next, we implement a constructor function. This is usefull in this case since we want to compute the prefactor
+of the potential once before we start our simulation.
+In Rust we typically use `new` for the constructors' name.
+
+```rust
+impl Mie {
+    fn new(sigma: f64, epsilon: f64, n: f64, m: f64) -> Mie {
```
Luthaf: If the `Mie::new` function is enforcing some constraint, maybe the fields should not be `pub`lic

doc/src/advanced/potential.md
```diff
@@ -0,0 +1,254 @@
...
+        let n_3 = self.n - 3.0;
+        let m_3 = self.m - 3.0;
+        let repulsive = f64::powf(sigma_rc, n_3);
+        let attractive = f64::powf(sigma_rc, m_3);
+        - self.prefac * self.sigma.powi(3) * (repulsive/n_3 - attactive/m_3)
+    }
+
+    fn tail_virial(&self, cutoff: f64) -> f64 {
+        if self.m <= 3.0 {
+            panic!("No tail correction possible for Mie potential with an exponent lower than 3.0")
```
Luthaf: In other long ranged potentials, we return from tail_energy/tail_virial instead of panicking.

Luthaf:
This looks very nice!

I have a suggestion about the organisation of this tutorial. Lumol is written such as it can be extended *from the outside*; which means the users do not need to write code in this repository to have it work with the rest of the engine. The tutorial could be written to use this property: just create a new cargo project, make it depend on the lumol crate, and implement the potential here.

This can have two benefits: 

- the tutorial can be made into a real repository that people can clone, and that can be tested with respect to lumol changes. 
- the tutorial does not need to deal with "extra work" that is needed for inclusion in this repository, but not when one just want to make the code work.

Your tutorial outline could then be splitted in two parts: the basics with

> Introduction
> Prerequisites
> Building the struct
> impl Potential
> impl PairPotential

and then the extra, only needed to include code in the main repository

> add API documentation
> add unit tests
> add parser for input files
> add documentation in the manual

The only issue is until we get a plugin system (#45), users needs to write there own `main` function if they want to use any newly developed potential/MC move/propagator/...

g-bauer:
Thanks for the feedback!

> The tutorial could be written to use this property: just create a new cargo project, make it depend on the lumol crate, and implement the potential here.

That's a good idea, I reorganize the tutorial accordingly. Would you still have this as part of the manual (I would, see the 2nd part about docs and input) or would we treat tutorials of this kind differently? Where would you put the tutorial repo?

I'll keep on pushing to this PR - it already includes all codes necessary to put them into a separate repo so this will be a quick change in the end.

Luthaf:
Concerning organisation, I think we can have everything in this repo for now, so the text in the manual, and the code in a tutorials directory. If it gets annoying, we can still move it to a separated repo later.

g-bauer:
Added first part of the tutorial. I also added a `tutorials` directory in the root with a separate package that implements the discussed material as well as a small sample simulation setup to test the implementation.

I somehow screwed up the git history (sorry).

doc/src/advanced/custom_potential/intro.md
```diff
@@ -0,0 +1,64 @@
+# Creating your own potential function
+
+In this tutorial we will teach you how to use your own potential function with Lumol.
+This tutorial consists of two parts.
+In the first part, we will implement the logic in a separate project using Lumol as an external library.
+The second part describes the neccessary steps to implement the logic into Lumol `core`.
```
Luthaf: into Lumol `core` => into Lumol core package, called `lumol-core`

Luthaf: into Lumol `core` => into Lumol core package called `lumol-core`

doc/src/advanced/custom_potential/intro.md
```diff
@@ -0,0 +1,64 @@
...
+
+A potential is a function that describes the energy and force between interaction sites.
+In Lumol we differentiate between two types of potentials: First, there are potential functions that need the global state of the system, *i.e.* all positions, as input. The Ewald summation which is used to compute electrostatic interactions is an example for this kind of potential.
+In Lumol we call them `GlobalPotential`s.
```
Luthaf: In Lumol we call them ... => We call them ...

doc/src/advanced/custom_potential/intro.md
```diff
@@ -0,0 +1,64 @@
...
+There are plenty of potentials falling into this category, hence in Lumol we simply call them `Potential`s.
+
+> There are two kind of potentials. A `GlobalPotential` takes the global systems' state as input
+and a `Potential` takes a single scalar value (distance, angle, ...) as input.
+
+In this tutorial we will focus on the implementation of the latter.
+More specific, we will implement a potential to compute van-der-Waals interactions between pairs of particles.
+We will make liberal use of the API documentation of both the Rust standard library as well as the Lumol API.
+Please note that we will point to other references, such as the Rust book, concerning general Rust concepts to keep this tutorial brief.
+If you have questions concerning Rust or Lumol, please don't hesitate to file an issue on github or join the discussion on our gitter.
```
Luthaf: maybe add the links to gitter and github here

tutorials/.gitignore
```diff
@@ -0,0 +1,3 @@
+/target/
+**/*.rs.bk
```
Luthaf: Is this the standard gitignore created by cargo new?

tutorials/mie_potential/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "mie_potential"
+version = "0.1.0"
```
Luthaf: I would use version = "0.0.0" here, and add a `publish = false` to ensure this is never published to crates.io

tutorials/mie_potential/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    let mut pairs = PairInteraction::new(Box::new(mie), 3.0 * 3.405);
+    // use tail corrections to account for our truncation
+    pairs.enable_tail_corrections();
+    // finally add this interaction to the system for Argon atoms
+    system.add_pair_potential("Ar", "Ar", pairs);
+    system
+}
+
+// This function runs a small Monte-Carlo simulation for Argon using a Mie potential.
+fn main() {
```
Luthaf: I would go for a simpler main, using the existing `Input` and just adding the Mie potential:

```rust
fn main() {
    let input = Input::new("some.path.toml").unwrap().read().unwrap();
    
    let mut system = input.system;
    let mut simulation = input.simulation;

    // ...
    system.add_pair_potential("Ar", "Ar", pairs);

    simulation.run(&mut system, input.nsteps);
}
```

This allow to not show too many things at once, and make it easier to introduce the second part later (the code in `lumol-input` is doing the same as the code in main, no more).

tutorials/mie_potential/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "mie_potential"
```
Luthaf: I would call this package `lumol-tutorial-potential`

doc/src/advanced/custom_potential/external.md
```diff
@@ -0,0 +1,205 @@
+# Part I: Implementation of the logic in a separate package
+
+We start by creating a new package using `cargo`:
+
+```
+cargo new mie_potential
```
Luthaf: Same as below, I would call the crate lumol-tutorial-potential

doc/src/advanced/custom_potential/external.md
```diff
@@ -0,0 +1,205 @@
...
+
+where $x$ denotes the distance between two interaction sites $i, j$, with $x = x_{ij} = | \mathbf{r}_j - \mathbf{r}_i |$. The parameters of the potential are
+
+- $n, m$ the repulsive and attractive exponents, respectively,
+- $\epsilon$ the energetic paramater,
+- $\sigma$ the particle diameter or structural parameter.
+
+We start by defining the `struct` for our potential. Add the following lines to `lib.rs`:
+
+```rust
```
Luthaf: Do you think it would be possible to directly include the relevant part of the file in `/tutorials` ? This would make it easier to prevent the code from bit rotting.

doc/src/advanced/custom_potential/external.md
```diff
@@ -0,0 +1,205 @@
...
+    /// Distance constant
+    sigma: f64,
+    /// Energy constant
+    epsilon: f64,
+    /// Exponent of repulsive contribution
+    n: f64,
+    /// Exponent of attractive contribution
+    m: f64,
+    /// Energetic prefactor computed from the exponents and epsilon
+    prefac: f64,
```
Luthaf: why not `prefactor`?

doc/src/advanced/custom_potential/external.md
```diff
@@ -0,0 +1,205 @@
...
+```rust
+pub trait PairPotential: Potential + BoxClonePair {
+    fn tail_energy(&self, cutoff: f64) -> f64;
+    fn tail_virial(&self, cutoff: f64) -> f64;
+
+    fn virial(&self, r: &Vector3D) -> Matrix3 { ... }
+}
+```
+
+First, we can see that `PairPotential` enforces the implementation of `Potential` which is denoted by `pub trait PairPotential: Potential ...` (we ignore `BoxClonePair` for now).
```
Luthaf: Maybe we can add a comment saying that `BoxClonePair` is automatically implemented for us if we implement `PairPotential` manually?

doc/src/advanced/custom_potential/external.md
```diff
@@ -0,0 +1,205 @@
...
+        -self.prefac * self.sigma.powi(3) * (repulsive * self.n / n_3 - attractive * self.m / m_3)
+    }
+}
+```
+
+Note that we cannot correct every kind of energy function.
+In fact, the potential has to be a *short ranged* potential.
+For our Mie potential, both the exponents have to be larger than 3.0 else our potential will be *long ranged* and
+the integral that has to be solved to compute the tail corrections diverges.
+For now, we will simply return zero in this case.
```
Luthaf: The "for now" looks strange here: returning 0 is the right thing to do if the potential is not short-ranged

Luthaf:
> I somehow screwed up the git history (sorry).

No, this does not look like so =)

> Not sure how long it will take me to write the second part. Maybe it makes sense to merge this as is.

Yeah, we can merge this as it is, and keep the advanced part of the tutorial for later.

----

Can you also add a test in .travis.yml that check that the code compiles (you don't need to run the simulation, just build everything with `cargo build`), and that it keeps compiling even when we will change other things?

doc/src/advanced/custom_potential/external.md
g-bauer: I'm note sure what you mean here. How would you directly include that part of the file?

Luthaf: Ok, I checked and this is not possible with markdown. So let's keep it this way for now.

I do this with sphinx: you have example files somewhere, and then you include them in your documentation. (see [here](https://github.com/chemfiles/chemfiles.py/commit/0576e9f5bb03a86a2e4554bb81445e8e27ff77db#diff-f1747d293233dbb0c4a0161f196a1917R29) for example). This allow to ensure that the documentation and the actual files stays in sync.

g-bauer: Ah I see, yes i do that in sphinx too.

.travis.yml
```diff
@@ -29,8 +29,12 @@ script:
     # Run unit tests and doc tests in debug mode
     - cargo test -p lumol-core
     - cargo test -p lumol-input
+    # Check if tutorials compile
+    - cd
```
Luthaf: Looks like a previous attempt leftover, but this might break the build, as the code is not in $HOME.

.travis.yml
```diff
@@ -29,8 +29,12 @@ script:
...
     # Check that benchmarks still compile
     - cargo bench --no-run
+    # Check if tutorials compile
+    - ./scripts/ci/compile-tutorials.py
```
Luthaf: I would rather explicitly do `cd tutorial/lumol_potential_tutorial && cargo build` here. We can craft a script for this later!

You might need to do a `cd $TRAVIS_BUILD_DIR` after that.

g-bauer: Sure, that was my first attempt :)

Luthaf: There is another way to do this, in one line: `cargo build --manifest-path=tutorial/lumol_potential_tutorial/Cargo.toml`.

But if you prefer the python script, you'll need to `chmod +x` it. There is a permission denied error on Travis.

g-bauer:
I'm done with this part.

Luthaf:
And thank you again for doing this: having a good documentation is essential, especially for open source software!

PR #199, opened 2017-10-09
Remove global translation and rotation on velocities initialization

Luthaf:
This prevent the flying ice-cube effect when the random velocity is not 100% isotropic

### DIFF
--- a/src/core/src/sim/md/controls.rs
+++ b/src/core/src/sim/md/controls.rs
@@ -109,7 +109,6 @@ impl Thermostat for BerendsenThermostat {}
 /******************************************************************************/
 
 impl<T> Control for Alternator<T> where T: Control {
-
     fn control(&mut self, system: &mut System) {
         if self.can_run() {
             self.as_mut().control(system)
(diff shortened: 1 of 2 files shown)


Luthaf:
@g-bauer, this is working now!

g-bauer:
I am wondering if it is necessary (or even wished) to have this utility inside the `scale` function. The very same function is used within the [velocity rescaling](https://github.com/lumol-org/lumol/blob/master/src/core/src/sim/md/controls.rs#L65) thermostat so that one effectively constrains the center of mass by using the thermostat.

Maybe we can use the control algorithms inside the initialization routines for the different methods to create velocities? Then, the user can control how/when center of mass and rotation is restrained using the control algorithms.

For some methods (to compute free energies of solids for example) it is vital that no center of mass control is used.

What do you think?

Luthaf:
> I am wondering if it is necessary (or even wished) to have this utility inside the scale function. The very same function is used within the velocity rescaling thermostat so that one effectively constrains the center of mass by using the thermostat.

Whoops  ...  Good catch! This was not intentional, I'll move the code inside the velocities initialization routines.

Issue #79, opened 2017-01-15
Clarify how to compute ideal gas contribution to pressure


g-bauer:
This issue was discussed in #58 and targets computation of the ideal gas pressure computation:

```rust
impl Compute for PressureAtTemperature {
    type Output = f64;
    fn compute(&self, system: &System) -> f64 {
        :
        let natoms = system.size() as f64;
        return natoms * K_BOLTZMANN * self.temperature / volume + virial / (3.0 * volume);
             //  ^--- natoms, degrees of freedom, number of molecules?
    }
}
```

Luthaf:
So I am pretty sure this should be the number of degrees of freedom. It is what [LAMMPS](http://lammps.sandia.gov/doc/compute_pressure.html) declare using in it's documentation, and it is what [RASPA](https://github.com/numat/RASPA2/blob/21a8c3f51c7c6247eaa7523f3333edb6fd2baea7/src/thermo_baro_stats.c#L201) (another MD and MC capable software) uses for both MC and MD computation.

It also goes to the number of atoms in fully flexible systems, and to the number of molecules in rigid molecules system.

Finally, in the [Allen and Tildesley book](https://www.amazon.com/Computer-Simulation-Liquids-Science-Publications/dp/0198556454), they derive the instantaneous pressure using the instantaneous temperature (eq 2.55), and the instantaneous temperature uses the number of degrees of freedom (eq 2.50).

You can find the [two corresponding pages from the book](https://github.com/lumol-org/lumol/files/814155/compsimliq.pdf), containing the whole derivation.

Luthaf:
Proposition to solve this: we can add a `Molecule::degree_of_freedom() -> usize` function, that is used to get the number of degree of freedom for each molecule, from 3 if the molecule is rigid to `3*natoms` if the molecule is flexible, going through `3*natoms - n_constraints` for constrained MD (RATTLE/SHAKE algorithms).

I think the main issue here is that we don't know whether a molecule is simulated a rigid or as flexible. We could start by assuming that MC means rigid and MD means flexible, but this would break quickly as we add more functionalities.

If/when we implement #137, we can have a nice way to resolve this. We can add a `rigid` key (`true/false/[list of constraints]`) in the molecule section, and use this in the `Molecule::degree_of_freedom` function.

What do you think?

g-bauer:
> I think the main issue here is that we don't know whether a molecule is simulated a rigid or as flexible. 

[...]

PR #204, opened 2017-10-15
Fix Resize move and add a simple NPT test

### LINKED ISSUE #203: Possibly bug in `resize` MC move
There is possibly a bug in the `resize` MC move (in [`prepare`](https://github.com/lumol-org/lumol/blob/master/src/core/src/sim/mc/moves/resize.rs#L83-L95)). I think we introduced it when we separated `configuration` and `system`.

This is mainly a remainder for me to have another look at it and to figure out a good integration test for NPT.

Luthaf:
Fix #203 

I only added test for an Argon system, molecular system will need #79 to be resolved first.


Luthaf:
And thank you very much for finding the issue here @g-bauer, this one was really bothering me, but I could not find the origin of the problem.

src/core/src/sim/mc/moves/resize.rs
```diff
@@ -81,13 +81,13 @@ impl MCMove for Resize {
...
-            // so we compute the translation vector of the center-of-mass
-            // (com) of a molecule and apply it to all its particles.
-            // Note that to do this, the com of a molecule *always* has
-            // to reside inside the simulation cell.
+            // We don't want to change the intramolecular distances so we
+            // compute the translation vector of the center-of-mass (com) of a
+            // molecule and apply it to all its particles. Note that to do
+            // this, the com of a molecule *always* has to reside inside the
+            // simulation cell.
             let old_com = system.molecule_com(mi);
```
g-bauer: It doesn't make a difference, but I think it would be more clear to use
```rust
let old_com = self.previous.molecule_com(mi);
```
here. Just a nitpick, we can also leave it as is.

g-bauer:
Thanks for the fast fix. Maybe you can comment on my suggestion and I'll merge.

src/core/src/sim/mc/moves/resize.rs
Luthaf: Yes, this seems more clear =)

g-bauer:
Another "unknown error". Do you want to rerun the test? Else I'd merge.

Luthaf:
I restarted the build.

Just wondering, do you have the rights to restart jobs too?

g-bauer:
Yes, I just wasn't logged in. Next time I can restart the build myself.

PR #210, opened 2017-10-19
Implement Gaussian potential

g-bauer:
This PR is an implementation of a Gaussian potential.

# Todo
- [x]  Add input manual entry (when we know where).
- [x]  Add tests for `lumol-input`.

Edit: Sorry for the formatting changes, this is due to `rustfmt` on save.

### DIFF
--- a/doc/reference/systems/interactions/potentials.rst
+++ b/doc/reference/systems/interactions/potentials.rst
@@ -186,3 +186,25 @@ units:
     [[pairs]]
     atoms = ["A", "B"]
     morse = {depth = "40 kJ/mol", a = "2.0 rad^-1", x0 = "109.7 deg"}
+
+
+Gaussian potential
+------------------
+
+This potential is usually used to describe energy wells and is expressed as:
+
+.. math::
+
+    V(r) = -a \exp(-b r^2)
+
+The potential type keyword is ``gaussian``, and the parameters ``a`` (well depth)
+and ``b`` (well width) should be provided as strings.
+
+.. code::
+
+    [[pairs]]
+    atoms = ["A", "B"]
+    gaussian = {a = "8.0 kJ/mol", b = "0.2 A^-2"}
+
+.. caution::
+    ``b`` has to be positive
--- a/src/core/src/energy/functions.rs
+++ b/src/core/src/energy/functions.rs
@@ -1,6 +1,7 @@
 // Lumol, an extensible molecular simulation engine
 // Copyright (C) Lumol's contributors — BSD license
 
+use std::f64::consts::PI;
 use math::*;
 use energy::Potential;
 use energy::{PairPotential, BondPotential, AnglePotential, DihedralPotential};
@@ -27,13 +28,21 @@ use energy::{PairPotential, BondPotential, AnglePotential, DihedralPotential};
 #[derive(Clone, Copy)]
 pub struct NullPotential;
 impl Potential for NullPotential {
-    fn energy(&self, _: f64) -> f64 {0.0}
-    fn force(&self, _: f64) -> f64 {0.0}
+    fn energy(&self, _: f64) -> f64 {
+        0.0
+    }
+    fn force(&self, _: f64) -> f64 {
+        0.0
+    }
 }
 
 impl PairPotential for NullPotential {
-    fn tail_energy(&self, _: f64) -> f64 {0.0}
-    fn tail_virial(&self, _: f64) -> f64 {0.0}
+    fn tail_energy(&self, _: f64) -> f64 {
+        0.0
+    }
+    fn tail_virial(&self, _: f64) -> f64 {
+        0.0
+    }
 }
 impl BondPotential for NullPotential {}
 impl AnglePotential for NullPotential {}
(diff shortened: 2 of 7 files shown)


src/core/src/energy/functions.rs
```diff
@@ -404,14 +421,61 @@ impl Potential for MorsePotential {
...
+/// use lumol_core::energy::Potential;
+/// use lumol_core::energy::Gaussian;
+///
+/// let potential = Gaussian{a: 8.0, b: 0.5};
+/// assert_eq!(potential.energy(0.0), -8.0);
+/// assert_eq!(potential.force(0.0), 0.0);
+/// ```
+#[derive(Clone, Copy)]
+pub struct Gaussian {
+    /// Distance constant of the Lennard-Jones potential
```
Luthaf: Should be Gaussian potential

src/core/src/energy/functions.rs
```diff
@@ -404,14 +421,61 @@ impl Potential for MorsePotential {
...
+    pub a: f64,
+    /// Energy constant of the Lennard-Jones potential
```
Luthaf: Same

src/core/src/energy/functions.rs
```diff
@@ -404,14 +421,61 @@ impl Potential for MorsePotential {
...
+    }
+
+    fn force(&self, r: f64) -> f64 {
+        2.0 * self.b * r * self.energy(r)
+    }
+}
+
+impl PairPotential for Gaussian {
+    fn tail_energy(&self, _: f64) -> f64 {
+        0.0
```
Luthaf: Is this temporary? It feels like this potential is not a long-ranged one. But I've never encountered it, so I don't know what it is used for.

src/core/src/energy/functions.rs
g-bauer: Ooops, nice catch, thanks.

src/core/src/energy/functions.rs
g-bauer: Good question :) For my use case the cut off distance is typically way larger than the potential's width. But you are right, I should implement it anyways.

Luthaf: Just out of curiosity, what do you use this potential for?

g-bauer: As an "external field". I use it to draw molecules to their reference positions in a crystalline phase. I want to use Lumol to compute free energies of solids. If you are interested in the method, I can give you more details.

Luthaf: Yes, I am always interested! Will you then to implement another propagator for this, or can you re-use the existing ones?

g-bauer: I might invest some time to bring basic free energy functionality to Lumol (lambda scaling) to make these computations more convenient.

g-bauer: As a first step I use a simple MC algorithm where the particles that interact via Gaussians are frozen in place (no translation). But I think that your idea of an external field could also work.

g-bauer:
There is both a check in the `new` function as well as the input parser for correct input. I also named the bad example for input test `gaussian-...` because I always have to go through multiple test cases to copy one I can reuse. We may want to give the test files more descriptive names 😄

doc/reference/systems/interactions/potentials.rst
```diff
@@ -186,3 +186,25 @@ units:
...
+
+
+Gaussian potential
+------------------
+
+This potential is usually used to describe energy wells and is expressed as:
+
+.. math::
+
+    V(r) = -a \exp(-b r^2)
```
Luthaf: What do you think of using another definition, namely `V = - depth exp(- (r / sigma)^2)` ? It feels clearer to me. We could also use epsilon instead of depth, to be closer to the LJ definition.

g-bauer: I went with what I found in papers that use this potential and the [docs of LAMMPS](http://lammps.sandia.gov/doc/pair_gauss.html). We can change it though if you find other expressions more suitable.

src/core/src/energy/functions.rs
```diff
@@ -404,14 +422,73 @@ impl Potential for MorsePotential {
...
+pub struct Gaussian {
+    /// Depth of the Gaussian potential
+    a: f64,
+    /// Width of the Gaussian potential
+    b: f64,
+}
+
+impl Gaussian {
+    pub fn new(a: f64, b: f64) -> Gaussian {
+        if a <= 0.0 || b <= 0.0 {
```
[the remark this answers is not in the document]
g-bauer: Never seen it, but yes it would work - only `b` has to be positive.

doc/reference/systems/interactions/potentials.rst
Luthaf: If they are previous occurrence of the potential with this form, it's fine for me =)

g-bauer:
Should I squash the commits?

Luthaf:
Yes please =)

Luthaf:
Only beta is failing, and I could not reproduce the failure locally, even with the official docker image from Travis. So I propose we merge this and other PR, and I'll try to debug the issue further this weekend.

g-bauer:
Sure! Do you merge?

PR #216, opened 2017-10-28
Use dist:precise on Travis

Luthaf:


Luthaf:
So it looks like process get killed on Travis with `dist: trusty` if they take too much memory. We where somehow hitting this with the latest beta in debug mode. Switching to `dist: precise` make it works!

PR #223, opened 2017-11-09
fixed typo which caused cargo not to run

g-bauer:


Luthaf:
Travis is failling with a DNS error (could not resolve github.com), I am restarting it

PR #224, opened 2017-11-18
Setup and run rustfmt

Luthaf:
This PR mainly setup and run rustfmt on the whole repo. Two other commits (2a6cb38 and 96d6e3a) do minor API changes that later improve the overall formatting.

I used to be pretty reluctant to use anything like a code formatter, but I recently changed my mind =). This will help when working with multiple people on the same codebase to maintain readability.

Future things we may want to do or not (what are your opinions on these?):

- Point to rustfmt in Contributing.md, asking contributors to reformat before submitting;
- Run rustfmt on Travis, failing the build if the code is not properly formatted.

I am a bit hesitant to do this now, as rustfmt is not stable at all and requires a nightly compiler to install.


g-bauer:
I am using rustfmt for quite a while and I like it - so I think it is a good idea to add it. I am not sure if I'd let the build fail if there are formatting issues though (at least for now) but if we decide to use it we should add it to Contributing.md.

Luthaf:
Ok, I added a few words about rustfmt in Contributing.md =)

g-bauer:
Can I merge?

Luthaf:
Yes, go for it!

If the PR is not marked as 'WIP', you are the one deciding to merge it =)

PR #225, opened 2017-11-19
Get the number of degrees of freedom from the propagator

Luthaf:
Here is my take on #79 !

~~The diff is pretty bad now, because this is based on #224. Once it is merged, I'll rebase here.~~

The idea here is to ask the propagator how many degrees of freedom are simulated, and then use this to compute the pressure/stress/temperature of the system.

This is marked as WIP because I want to add more MC tests using it and check that the original issue is actually solved.

### DIFF
--- a/src/core/src/sim/mc/mod.rs
+++ b/src/core/src/sim/mc/mod.rs
@@ -6,5 +6,5 @@ mod monte_carlo;
 pub use self::monte_carlo::{MonteCarlo, MoveCounter};
 
 mod moves;
+pub use self::moves::{MCDegreeOfFreedom, MCMove};
 pub use self::moves::{Resize, Rotate, Translate};
-pub use self::moves::MCMove;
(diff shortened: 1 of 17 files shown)


Luthaf:
So, more explanations on what is going on here. 

The main commits are 999d4a6, 3d5841b and bd09417. The number of simulated degrees of freedom are taken from the propagator as the `DegreesOfFreedom` enum, and the corresponding `MCDeegresOfFreedom` enum for MC Moves.

The other commits are Yak shaving: I needed the molecular composition (
cb22b56 ) to send warnings [here](https://github.com/lumol-org/lumol/pull/225/commits/3d5841be9786321c71710e6b468ca4b4dab5504d#diff-5e1aaf4a5e540529065afa8c35467a94R169), but then computing the molecular types every time starts to be expensive, so I added a cache for them (790c982 ), and fixed a bug in the process (89cb3dc).

Concerning the Composition, maybe we don't want to put atomic and molecular composition together: we never need both of them at the same time. So it might make sense to obtain them from different function, to remove a bit of work in tight loops.

PR #234, opened 2018-03-17
Add WCA

gjepson:
New WCA code examples, pair potentials. All tests passed.


g-bauer:
Hey! Thanks for the PR, I'll have a look at it later today.

g-bauer:
Small nitpick: Not sure if rust does this for us, but you could replace `f64::powi(s6, 2)` with `s6 * s6` within the energy and force computation.

I'm not sure I understand the correction terms. Would it make sense to use the regular formulation of tail corrections but using `2^1/6 sigma` as upper limit instead of infinity (see [here](http://lumol.org/lumol/latest/book/input/systems/interactions/non_bonded.html#tail-correction))? We could use such a correction when `rc < 2^1/6 sigma` and zero if `rc > 2^1/6 sigma`, or am I missing something?

Do you need help making the potential accessible from input files?

gjepson:
s6 * s6 works fine as well. I used f64::powi(s6, 2) because it was used in the Lennard-Jones potential implementation. Is there a benefit to using one over the other? 

Regarding the cutoffs, I added a new commit which should correct the correction terms. If I'm thinking of this correctly, WCA converges, so this should yield 0 if the cutoff is beyond 2^1/6 sigma, and panic otherwise. 

I'll check through the documentation regarding potential accessibility from input files and let you know if I have any questions. 

Travis hasn't passed at the time of commenting, so it this may well be moot if it doesn't pass.

Luthaf:
> s6 * s6 works fine as well. I used f64::powi(s6, 2) because it was used in the Lennard-Jones potential implementation. Is there a benefit to using one over the other?

`s6 * s6` should be faster, as `powi` might can a function in libm. Expect from that, there should be no difference.

src/core/src/energy/functions.rs
```diff
@@ -104,6 +104,69 @@ impl PairPotential for LennardJones {
...
+impl PairPotential for WeeksChandlerAndersen {
+    fn tail_energy(&self, cutoff: f64) -> f64 {
+        if cutoff < f64::powf(2., 1./6.) * self.sigma {
+            panic!("Error: WCA Cutoff too small: {} < {}", cutoff, f64::powf(2., 1./6.) * self.sigma);
+        }
+        0.0
+    }
+    fn tail_virial(&self, cutoff: f64) -> f64 {
+        if cutoff < f64::powf(2., 1./6.) * self.sigma {
+            panic!("Error: WCA Cutoff too small: {} < {}", cutoff, f64::powf(2., 1./6.) * self.sigma);
```
Luthaf: Could you change the message to mention "tail corrections"? Something like `WCA cutoff ({}) is smaller than the tail corrections cutoff ({})`?

src/core/src/energy/functions.rs
```diff
@@ -104,6 +104,69 @@ impl PairPotential for LennardJones {
...
+pub struct WeeksChandlerAndersen {
+    /// Distance constant for the WeeksChandlerAndersen Potential
+    pub sigma: f64,
+    /// Energy Constant for the WeeksChandlerAndersen Potential
+    pub epsilon: f64,
+}
+
+impl Potential for WeeksChandlerAndersen {
+    fn energy(&self, r: f64) -> f64 {
+        if r < f64::powf(2., 1./6.) * self.sigma {
```
Luthaf: For better efficiency, the `f64::powf(2., 1./6.) * self.sigma` value could be stored inside the `WeeksChandlerAndersen` struct, there is no need to recompute it every time. But that is not required.

Luthaf:
So what is missing here? I'd like to get this to go forward =)

I think this could be merged with some additional regressions tests; something like this: 

https://github.com/lumol-org/lumol/blob/0f6efbd38b0bf132af2f912c0b233e5a8829bf02/src/core/src/energy/functions.rs#L643-L658

Testing some specific values for energy and forces, and checking that the force is indeed the derivative of the energy using finite differences.

Support for WCA in input can come later.

Luthaf:
Closing this due to inactivity. Feel free to create a new PR if you are still interested!

PR #235, opened 2018-03-24
Add mie

g-bauer:
This PR adds the Mie potential. Based on this PR, I'll finish the tutorial on how to add pair potentials to Lumol.

I created separate files for the input file tests - makes searching things a lot easier. Hope that's ok.

PR #236, opened 2018-03-26
Input cleanup

Luthaf:
This PR is a mismatch of cleanup tasks of the input code and test cases. Each commit should be self-contained and explanatory. Each commit/change is open to discussion!


g-bauer:
Looks good to me, especially that we can cover all cases for potentials within a single file. I merge and rebase #235.

Luthaf:
thanks!

PR #241, opened 2018-04-06
Dependencies update

Luthaf:


Luthaf:
(Travis failure is the same as #240, with the same fix bd9aa299f463ee18cf157fd75b95fedbe9b099bd)

